PR #34, opened 2026-01-16
Add a plugin to the instruction-plan package to send as a single transaction

mcintyre94:
This PR adds a new `sendTransaction` plugin to the instruction-plans package. This uses the same internals as `send`, but errors if the planned `TransactionPlan` is not a single transaction message. It then unpacks the single transaction plan result and returns just a `TransactionPlanResultStatus`.

This makes the result much easier for consumers to handle, as they don't need to consider results of multiple transactions.

I've updated the example from this PR stack to demonstrate this simplification

Draft, TODOs:

- Need to figure out what errors from kit-plugins should look like. Should we just reserve some space in the kit package, since that uses `@solana/errors`? It could be confusing to have errors across multiple packages, though that's also inevitable with other packages. It's also quite complex to replicate `@solana/errors`
    - Also note that `SOLANA_ERROR__TRANSACTION__FAILED_WHEN_SIMULATING_TO_ESTIMATE_COMPUTE_LIMIT` is in Kit, but only thrown by compute-budget
- If we have asserted that a `TransactionPlan` has `kind === "single"`, then can we update the types in Kit to return a `TransactionPlanResult` with the kind known to be `single`? If so this can be simplified a bit. Related Kit issue: https://github.com/anza-xyz/kit/issues/1209

### DIFF
--- a/examples/transfer-lamports/src/index.ts
+++ b/examples/transfer-lamports/src/index.ts
@@ -18,7 +18,7 @@ const client = await createDefaultLocalhostRpcClient();
 const destinationAccountAddress = (await generateKeyPairSigner()).address;
 
 try {
-    const result = await client.send(
+    const result = await client.sendTransaction(
         getTransferSolInstruction({
             amount: lamports(10_000_000n),
             // 0.01 SOL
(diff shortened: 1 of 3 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/34?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#34** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/34?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/34?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#33** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/33?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #35, opened 2026-01-16
Unwrap the nested simulation failure in transaction plan error

### LINKED ISSUE #25: Improve error handling for `sendInstructionPlans` when the transaction fails to simulate
Probably the most common reason for a transaction to fail during development is that something in the instructions is wrong, eg trying to send more SOL than the account has. In other words, `sendTransaction` or `simulateTransaction` will return a `TransactionError`.

This error becomes extremely obfuscated by our `client.send` function:

```
SolanaError: The provided transaction plan failed to execute. See the `transactionPlanResult` attribute and the `cause` error for more details.
    at file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana+instruction-plans@5.3.0_typescript@5.9.3/node_modules/@solana/instruction-plans/dist/index.node.mjs:225:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Object.send (file:///home/sol/projects/kit-plugins/packages/kit-plugin-instuction-plan/dist/index.node.mjs:22:14)
    at async file:///home/sol/projects/kit-plugins/examples/transfer-lamports/src/index.ts:20:1 {
  cause: SolanaError: Transaction failed when it was simulated in order to estimate the compute unit consumption. The compute unit estimate provided is for a transaction that failed when simulated and may not be representative of the compute units this transaction would consume if successful. Inspect the `cause` property of this error to learn more
      at simulateTransactionAndGetConsumedUnits (file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana-program+compute-budget@0.11.0_@solana+kit@5.3.0_typescript@5.9.3_/node_modules/@solana-program/compute-budget/dist/src/index.mjs:382:14)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async estimateComputeUnitLimit (file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana-program+compute-budget@0.11.0_@solana+kit@5.3.0_typescript@5.9.3_/node_modules/@solana-program/compute-budget/dist/src/index.mjs:354:10)
      at async estimateComputeUnitLimitFactoryFunction (file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana-program+compute-budget@0.11.0_@solana+kit@5.3.0_typescript@5.9.3_/node_modules/@solana-program/compute-budget/dist/src/index.mjs:409:12)
      at async file:///home/sol/projects/kit-plugins/packages/kit-plugin-instuction-plan/dist/index.node.mjs:86:36
      at async fn (file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana-program+compute-budget@0.11.0_@solana+kit@5.3.0_typescript@5.9.3_/node_modules/@solana-program/compute-budget/dist/src/index.mjs:339:7)
      at async file:///home/sol/projects/kit-plugins/packages/kit-plugin-instuction-plan/dist/index.node.mjs:113:25
      at async file:///home/sol/projects/kit-plugins/packages/kit-plugin-instuction-plan/dist/index.node.mjs:114:65
      at async file:///home/sol/projects/kit-plugins/packages/kit-plugin-instuction-plan/dist/index.node.mjs:110:35
      at async traverseSingle (file:///home/sol/projects/kit-plugins/node_modules/.pnpm/@solana+instruction-plans@5.3.0_typescript@5.9.3/node_modules/@solana/instruction-plans/dist/index.node.mjs:263:20) {
    cause: { InstructionError: [Array] },
    context: { __code: 5663019, unitsConsumed: 300 }
  },
  context: { __code: 7618003 }
}
```

What's going on here is:

- the `transactionPlanExecutor` uses `estimateAndSetCULimit`
- that calls `simulateTransaction`, and gets back the `TransactionError`

The error structure and stacktrace make sense, but they expose the details of how we estimate CU limits and how we're using a transaction plan

Processing this error requires the developer to be aware of both errors and to handle the nesting. Just showing the happy path:

```ts
if (isSolanaError(e, SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN)) {
    if (isSolanaError(e.cause, SOLANA_ERROR__TRANSACTION__FAILED_WHEN_SIMULATING_TO_ESTIMATE_COMPUTE_LIMIT)) {
        const transactionPlanResult = e.context.transactionPlanResult as TransactionPlanResult;
        const summary = summarizeTransactionPlanResult(transactionPlanResult);
        const transactionMessage = summary.failedTransactions[0].message;

        const transactionError = e.cause.cause as TransactionError;
        const nestedError = getSolanaErrorFromTransactionError(transactionError);

        if (isSystemError(nestedError, transactionMessage)) {
            console.error(getSystemErrorMessage(nestedError.context.code));
            return;
        }
    }
}
```

Given that this is probably the most likely error case for `client.send`, I think we should probably handle it in `sendInstructionPlans`. And throw something easier to handle with just the nested `TransactionError` (or `getSolanaErrorFromTransactionError` for it) + the transaction message.

Note related issue in Kit: https://github.com/anza-xyz/kit/issues/1201

If we made the change suggested there then this error would move to the `failedTransactions` on the result of `client.send` instead of being a thrown error. In this case, it could make sense for us to instead post-process the result and unwrap this nested error structure before returning it.

mcintyre94:
This PR updates the instruction plan plugin to improve how we handle probably the most common error case: the transaction failing to simulate

When this happens, we get a `SOLANA_ERROR__TRANSACTION__FAILED_WHEN_SIMULATING_TO_ESTIMATE_COMPUTE_LIMIT` as the cause of a `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN`

The simulate error contains a `TransactionError`, which needs to be converted to a `SolanaError` in order to use helpers like `isSystemError`

With this PR, this unrolling is done for you, and the plugin throws a `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN` with the cause being the `SolanaError` for the nested `TransactionError`, and the `transactionPlanResult` being the same as the original error.

As can be seen in the example, the resulting error handling for apps becomes much simpler.

Drafts, TODOs:
- [ ] tests
- [ ] better error for the single transaction case. Here we shouldn't expose transaction plans, and we should extract the transaction message from the plan result. This will require a new error type and further simplify the single transaction example 

Closes #25

### DIFF
--- a/packages/kit-plugin-instruction-plan/src/core.ts
+++ b/packages/kit-plugin-instruction-plan/src/core.ts
@@ -12,6 +12,7 @@ import {
     TransactionPlan,
     TransactionPlanExecutor,
     TransactionPlanner,
+    TransactionPlanResult,
 } from '@solana/kit';
 
 /**
@@ -95,7 +96,9 @@ export function sendInstructionPlans() {
             const transactionPlan = await transactionPlanner(instructionPlan, innerConfig);
             config?.abortSignal?.throwIfAborted();
             const transactionPlanExecutor = config.transactionPlanExecutor ?? client.transactionPlanExecutor;
-            return await transactionPlanExecutor(transactionPlan, innerConfig).catch(unwrapSimulationError);
+            return await transactionPlanExecutor(transactionPlan, innerConfig)
+                .then(convertPlanResult)
+                .catch(rethrowWithConvertedSimulationError);
         },
     });
 }
@@ -127,7 +130,8 @@ export function sendTransaction() {
                 // `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN`:
                 // - Get the `transactionMessage` from the first failed transaction
                 // - Return `e.cause` + `transactionMessage` with a named SendTransactionError
-                .catch(unwrapSimulationError);
+                .then(convertPlanResult)
+                .catch(rethrowWithConvertedSimulationError);
             // TODO: can we tighten the types in Kit so that single transaction plan -> single transaction plan result? Does that hold?
             if (result.kind === 'single') {
                 return result.status;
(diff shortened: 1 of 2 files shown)


packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -143,3 +158,52 @@ function assertIsSingleTransactionPlan(
...
+                ...result,
+                plans: result.plans.map(convertPlanResult),
+            };
+        case 'parallel':
+            return {
+                ...result,
+                plans: result.plans.map(convertPlanResult),
+            };
+    }
+}
```
mcintyre94: Maybe also an argument for moving this into Kit, as a generic `mapFailed` function?

PR #45, opened 2026-01-23
Allow custom RPC configuration in localhost client

beeman:
This PR enables connecting the default localhost rpc client to a custom rpc and websocket url by exposing a new optional property `rpc`.

This is useful for connecting the client to a test validator running in a container [like this](https://github.com/beeman/testcontainers).

### DIFF
--- a/packages/kit-plugin-rpc/src/index.ts
+++ b/packages/kit-plugin-rpc/src/index.ts
@@ -65,6 +65,6 @@ export function rpc<TClusterUrl extends ClusterUrl>(
  *
  * @see {@link rpc}
  */
-export function localhostRpc() {
-    return rpc('http://127.0.0.1:8899', { url: 'ws://127.0.0.1:8900' });
+export function localhostRpc(url?: string, rpcSubscriptionsConfig?: { url: string }) {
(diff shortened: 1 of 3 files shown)


beeman:
Thanks for the review and comment @lorisleiva. 

I considered adding these properties but decided to try and implement a solution with the minimal amount of changes required - that's why I went with passing along `rpc`.

I'd love to have this use specific case supported through the plugins, so more than happy to update this PR.

Is your idea to pass it along like so? Or is there a better/easier way to do this?

```ts
// update the localhostRpc
export function localhostRpc(url?: string, rpcSubscriptionsConfig?: { url: string }) {
    return rpc(url ?? 'http://127.0.0.1:8899', rpcSubscriptionsConfig ?? { url: 'ws://127.0.0.1:8900' });
}
```

```ts
// update where it's used
export function createDefaultLocalhostRpcClient(
    config: { payer?: TransactionSigner; rpcSubscriptionsConfig?: { url: string }; url?: string } = {},
) {
    return createEmptyClient()
        .use(localhostRpc(config.url, config.rpcSubscriptionsConfig))
        .use(airdrop())
        .use(payerOrGeneratedPayer(config.payer))
        .use(defaultTransactionPlannerAndExecutorFromRpc())
        .use(sendTransactions());
}
```

PR #49, opened 2026-01-27
feat(litesvm): add transaction result assertions and type guards

amilz:
## Summary of Changes
Add utility functions for type-safe handling of LiteSVM transaction results including:
- Type guards: `isSuccessfulTransaction`, `isFailedTransaction`,` isSuccessfulSimulation`
- Assertions: `assertIsSuccessfulTransaction`, `assertIsFailedTransaction`, `assertIsSuccessfulSimulation`
- Type definitions: `SendTransactionResult`, `TransactionResult`, `SimulationResult`
- Unit tests

## Problem Solved
LiteSVM transaction methods (`sendTransaction`, `airdrop`, `simulateTransaction`, `getTransaction`) return union types that can be successful results, failures, or `null`. Without type guards, developers must write repetitive boilerplate to safely narrow these types before accessing result-specific methods like `computeUnitsConsumed()` or `err()`.  

Alternatively, this could go into LiteSVM sdk directly and reexported here, but I feel ammending your litesvm pr could risk slowing down that merge. WDYT @lorisleiva?

### DIFF
--- a/packages/kit-plugin-litesvm/src/assertions.ts
+++ b/packages/kit-plugin-litesvm/src/assertions.ts
@@ -1,0 +1,125 @@
+import type { FailedTransactionMetadata, TransactionMetadata } from '@loris-sandbox/litesvm-kit';
+import { SimulatedTransactionInfo } from '@loris-sandbox/litesvm-kit';
+
+export type { FailedTransactionMetadata, TransactionMetadata } from '@loris-sandbox/litesvm-kit';
+export { SimulatedTransactionInfo } from '@loris-sandbox/litesvm-kit';
+
+/** Result from `sendTransaction`. */
+export type SendTransactionResult = FailedTransactionMetadata | TransactionMetadata;
+
+/** Result from `airdrop` or `getTransaction` (can be `null`). */
+export type TransactionResult = FailedTransactionMetadata | TransactionMetadata | null;
+
+/** Result from `simulateTransaction`. */
+export type SimulationResult = FailedTransactionMetadata | SimulatedTransactionInfo;
+
+/**
+ * Checks if a transaction result is successful (`TransactionMetadata`).
+ *
+ * @example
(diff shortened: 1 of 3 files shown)


lorisleiva:
Hey @amilz, these helpers could be useful but my main concern is that they would be re-exported from `@solana/kit-plugins` and, without context, it wouldn't be clear that say `isSuccessfulTransaction` is related to the `LiteSVM` API.

IMO, a better approach is to expose the `client.svm` instance (for things like `addProgramFromFile` or `setAccount`) but we abstract it for sending and confirming transactions.

The `instruction-plans` plugin allows this because it creates a `transactionPlanner` and a `transactionPlanExecutor` given a `LiteSVM` instance. I'm actually about to tweak this plugin a little to improve its developer experience as we've made some new changes to `@solana/kit` itself to accommodate for this.

For instance, it has new type guards for `TransactionPlanResults` so you can do things like `assertIsSuccessfulTransactionPlanResult` ([PR](https://github.com/anza-xyz/kit/pull/1253), [PR](https://github.com/anza-xyz/kit/pull/1254)) and never have to worry about the fact that it was sent using an RPC or a LiteSVM instance.

We're also going to rename `client.send` to `client.sendTransactions` so we can add a new simpler `client.sendTransaction` helper that lets you focus on a single transaction result instead of a whole plan. In the end, you will be able to handle sending transactions like so:

```ts
const client = await createDefaultLiteSVMClient();
client.svm.setAccount(myTestAccount);
client.svm.addProgramFromFile(myProgramAddress, 'program.so');

const result = await client.sendTransaction([instructionA, instructionB]); // or pass an already prepared `TransactionMessage` directly.
console.log(result.status.signature);
```

What do you think?

amilz:
Sounds awesome @lorisleiva ! Happy to take a look/test it out if you want to tag me whenever that's up.

PR #57, opened 2026-01-28
Add `client.sendTransaction` to `sendTransactions` plugin

lorisleiva:
This PR adds a new `client.sendTransaction` helper in the `sendTransactions` plugin.

This helper ensures only a single transaction is executed, regardless of the instruction plan or instruction array provided. Should these instructions not fit in a single transaction, an error will be thrown before trying to execute it.

Note that I prematurely updated the README and docblocks to also mention the transaction message input so that I didn't have to revamp the documentation every five minutes but this change will be implemented in the next PRs of this stack.

### DIFF
--- a/.changeset/little-kings-wonder.md
+++ b/.changeset/little-kings-wonder.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-instruction-plan': minor
+---
+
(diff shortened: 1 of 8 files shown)


lorisleiva:
### Merge activity

* **Jan 28, 8:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/57).

PR #58, opened 2026-01-28
Add `TransactionMessage` support to `sendTransactions` helpers

lorisleiva:
This PR updates the `client.sendTransaction` and `client.sendTransactions` functions of the `sendTransactions` plugin such that it accepts direct `TransactionMessages` as input.

The former accepts a single `TransactionMessage & TransactionMessageWithFeePayer` whereas the latter also accepts an array of them.

Note that it would be nice to also support `TransactionPlans` to be passed directly but there is currently no convenient way to distinguish between `InstructionPlans` and `TransactionsPlans` (unless we check for their leaves recursively).

### DIFF
--- a/.changeset/fifty-dodos-take.md
+++ b/.changeset/fifty-dodos-take.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-instruction-plan': minor
+---
+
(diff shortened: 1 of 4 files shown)


mcintyre94 (approved):
Nice! Just one suggestion that could make this a little more flexible

mcintyre94 (approved):
Looks great, thankyou!

PR #61, opened 2026-01-29
Add a kit-plugins version of the transfer-lamports example from Kit

mcintyre94:
This PR adds an examples/ directory, which is used to store example scripts making use of kit-plugins.

It adds a first example based off Kit's transfer-lamports example. See for comparison: https://github.com/anza-xyz/kit/blob/main/examples/transfer-lamports/src/example.ts

It uses client.sendTransaction to automatically plan + execute based on just a getTransferSolInstruction.
In the success case, the app can just use result.status.signature to access the signature of the sent transaction.

It also demonstrates error handling. Thanks to our recent changes the app only needs to catch `SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN`, and receives any TransactionError already converted to a SolanaError - so it can use eg isSystemError directly.

It can trigger a failure state by being run with --fail, which will change the source of the transferSol instruction to a randomly generated address instead of the funded one created by createDefaultLocalhostRpcClient.

Example outputs:

```bash
$ pnpm start

> @solana/examples-transfer-lamports@ start /home/sol/projects/kit-plugins/examples/transfer-lamports
> node src/index.ts

Transaction successful! https://explorer.solana.com/tx/XaYkaYYDTKSg4YcVW89XRuToeBz8YifwHHw9M6vv2t5sVj6x3KXdpsFk8sPG6dU79RasnLeKkA9ZvCG5jjETBcD?cluster=custom&customUrl=http%3A%2F%2F127.0.0.1%3A8899
```

```bash
$ pnpm start --fail

> @solana/examples-transfer-lamports@ start /home/sol/projects/kit-plugins/examples/transfer-lamports
> node src/index.ts --fail

System Program Error: account does not have enough SOL to perform the operation
{
  instructionType: 'TransferSol',
  programAddress: '11111111111111111111111111111111',
  accounts: {
    source: {
      address: '9oFRLzGmNBG8YBQLPZbeJ539jrDTVAyhyaLBg9nKdQtT',
      role: 3,
      signer: [Object]
    },
    destination: {
      address: 'CE3CRz7xD1ZMnehZPpp5zAn1uXpZENbAP5oePNbTCQJc',
      role: 1
    }
  },
  data: { discriminator: 2, amount: 10000000n }
}
 ELIFECYCLE  Command failed with exit code 1.
```


lorisleiva (approved):
Great example!

Issue #68, opened 2026-02-06
kit-plugins package breaks browser builds by bundling litesvm


amilz:
### Problem

`@solana/kit-plugins` re-exports everything from all individual plugin packages (including `@solana/kit-plugin-litesvm`).

When you import only what you need:

```javascript
import { rpc, payer } from '@solana/kit-plugins';
```

Bundlers (Vite/esbuild/Rollup) still resolve the full dependency graph and pull in `@solana/kit-plugin-litesvm`, which includes a native `.node` binary (`litesvm-kit-darwin-arm64`). This causes browser build fails and a large dep set even when not using litesvm 

### Workaround/Proposed Fix

Import from individual packages works fine, eg:
```javascript
import { rpc } from '@solana/kit-plugin-rpc';
import { payer } from '@solana/kit-plugin-payer';
```

There are probably a few ways to fix this, but IMO since litesvm is a very specific testing-only environment/tool that should only ever get installed as a devDep:
- remove litesvm from @solana/kit-plugins
- include the default liteSVM client (which is very handy) inside the litesvm package 

fyi @lorisleiva

amilz:
Hey @lorisleiva -- thanks for #76. 
Doing some experimenting today on a remote server, and I think this also affects Node.js environments on Linux with glibc < 2.38, not just browsers. 

Environment:
• Debian 12 (bookworm), glibc 2.36
• Node.js v22.22.0
• Running inside Docker (common for CI, servers, and agent runtimes)
• Bun 1.3.9 exhibits identical behavior

### Reproduce 
Either of these crash immediately — even though we never call litesvm()
```ts
require('@solana/kit-plugins');
// or
import { createDefaultLocalhostRpcClient } from '@solana/kit-plugins';
```

```bash
Error: Cannot find native binding.
  [cause]: litesvm.linux-x64-gnu.node: undefined symbol: __isoc23_strtol
    code: 'ERR_DLOPEN_FAILED'

Stack: @loris-sandbox/litesvm-kit/dist/internal.js
     → @solana/kit-plugin-litesvm/dist/index.node.cjs
     → @solana/kit-plugins/dist/index.node.cjs
```

LMK what you think -- can open a separate issue as well.

lorisleiva:
Hey @amilz thanks for testing this out! I've got a few questions to help me understand the issue.

[...]

PR #76, opened 2026-02-11
Stub LiteSVM in browser/react-native builds

lorisleiva:
This PR replaces the content of the `@solana/kit-plugin-litesvm` package such that it returns a mock `litesvm` plugin that throws an error in Browser and React Native builds. That means, these builds are no longer bundling the `@loris-sandbox/litesvm-kit` dependency in the Browser which should fix issue #68.


mcintyre94 (approved):
Looks good!

PR #95, opened 2026-02-24
Fix node:fs side-effect import leaking into non-Node builds

lorisleiva:
This PR fixes a bug where `import 'fs'` was present in the browser and React Native build outputs of `@solana/kit-plugin-payer`, even though `payerFromFile` is guarded by `__NODEJS__` and its body is dead-code-eliminated. The root cause: tsup's esbuild stage removes the `readFileSync` binding but converts the import to a bare side-effect import (`import 'fs'`). The subsequent Rollup tree-shaking stage preserves it because it assumes all modules have side effects by default. The fix sets the `treeshake` option in the shared tsup config to a Rollup options object with a `moduleSideEffects` callback that returns `false` for `fs` and `path` modules, allowing Rollup to drop the import when no bindings are used.

### DIFF
--- a/.changeset/dry-foxes-rule.md
+++ b/.changeset/dry-foxes-rule.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-payer': patch
+---
+
+Fixed `node:fs` side-effect import leaking into browser and React Native builds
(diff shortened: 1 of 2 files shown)


lorisleiva:
### Merge activity

* **Feb 26, 8:33 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/95).

PR #96, opened 2026-02-24
Align kit-plugins package infrastructure with sub-packages

lorisleiva:
This PR removes unused IIFE build infrastructure from the umbrella package and shared tsup config. The jsdelivr, umd, and unpkg fields in @solana/kit-plugins's package.json pointed to dist/index.production.min.js, a file that was never produced since no build config included the iife format. The corresponding dead IIFE plumbing in the shared tsup base config (browserslist targets, minification branch, iife output extension, conditional sourcemaps) and the browserslist-to-esbuild dev dependency are also removed. The umbrella package's package.json now matches the sub-packages exactly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
         "@solana/prettier-config-solana": "0.0.6",
         "@types/node": "^25",
         "agadoo": "^3.0.0",
-        "browserslist-to-esbuild": "^2.1.1",
         "eslint": "^9.39.2",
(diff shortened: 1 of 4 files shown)


lorisleiva:
### Merge activity

* **Feb 26, 8:33 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/96).

PR #101, opened 2026-02-24
Add `skipPreflight` option to `defaultTransactionPlannerAndExecutorFromRpc`

lorisleiva:
This PR adds a `skipPreflight` option to `defaultTransactionPlannerAndExecutorFromRpc` and fixes the preflight behavior to be context-aware.

Previously, the executor always set `skipPreflight: true` when sending transactions, even when no simulation had been performed (explicit CU limit case), meaning users got no simulation at all. Now, preflight is only skipped when a CU estimation simulation was already performed. When the transaction has an explicit compute unit limit, preflight runs as the only simulation.

Setting `skipPreflight: true` changes the behavior in two ways: preflight is always skipped, and failed CU estimation simulations are recovered from by using the consumed units to set the CU limit — useful for debugging failed transactions in an explorer.

The CU estimation logic is split into two focused helpers — `needsComputeUnitEstimation` and `estimateAndSetComputeUnitLimit` — using `estimateComputeUnitLimitFactory` and `findSetComputeUnitLimitInstructionIndexAndUnits` from `@solana-program/compute-budget@0.14.0`.

### DIFF
--- a/.changeset/public-bags-guess.md
+++ b/.changeset/public-bags-guess.md
@@ -1,0 +1,13 @@
+---
+'@solana/kit-plugin-instruction-plan': minor
+---
+
+Add `skipPreflight` option to `defaultTransactionPlannerAndExecutorFromRpc`
+
+The executor now handles preflight simulation intelligently based on whether compute unit estimation was performed:
+
+- When estimation succeeds, preflight is skipped (avoiding a redundant second simulation).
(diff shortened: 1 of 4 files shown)


lorisleiva:
### Merge activity

* **Feb 26, 8:33 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/101).

PR #102, opened 2026-02-24
Move transaction planner/executor plugins to `kit-plugin-rpc` and `kit-plugin-litesvm`

lorisleiva:
This PR moves the RPC and LiteSVM transaction planner and executor plugins out of `@solana/kit-plugin-instruction-plan` into their respective `@solana/kit-plugin-rpc` and `@solana/kit-plugin-litesvm` packages. The combined `defaultTransactionPlannerAndExecutorFromRpc`/`defaultTransactionPlannerAndExecutorFromLitesvm` functions are replaced with separate `rpcTransactionPlanner()` + `rpcTransactionPlanExecutor()` and `litesvmTransactionPlanner()` + `litesvmTransactionPlanExecutor()` plugins. The `@solana/kit-client-rpc` and `@solana/kit-client-litesvm` packages are updated to use the new imports.


mcintyre94 (approved):
Nice! This feels much cleaner and would be much better for someone wanting to add eg another similar-to-litesvm client too.

PR #106, opened 2026-02-24
Expose typed `context.transactionMetadata` on the LiteSVM executor

lorisleiva:
This PR exposes a typed `context.transactionMetadata` property on the LiteSVM transaction plan executor result. The property contains the `TransactionMetadata` (or `FailedTransactionMetadata` on failure) returned by LiteSVM's `sendTransaction`, letting consumers inspect logs, compute units consumed, return data, and other metadata directly from the plan result. The executor context is now properly typed via `createTransactionPlanExecutor<{ transactionMetadata: ... }>`, and the metadata types are re-exported from the package. The executor also replaces its local `LiteSVM` duck type with the real type from `@loris-sandbox/litesvm-kit` via type-only imports (safe for browser builds since they're erased at compile time).

### DIFF
--- a/.changeset/dry-donuts-jog.md
+++ b/.changeset/dry-donuts-jog.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-litesvm': patch
+---
+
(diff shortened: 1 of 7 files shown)


lorisleiva:
### Merge activity

* **Feb 26, 8:33 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/106).

PR #173, opened 2026-03-31
Scaffold/design for the wallet plugin

mcintyre94:
This PR adds the scaffolding and design of the proposed wallet plugin, without yet implementing it

The plugin is designed to expose the interface expected by reactive frameworks, eg React's [useSyncExternalStore](https://claude.ai/chat/ea9e8f54-4d53-42fc-a3a9-1e6c7c1b5e37). 

## Example (Vanilla JS)

```ts
const client = createLocalClient().use(walletSigner{ chain: 'solana:localnet' })

// call render whenever something changes
client.wallet.subscribe(render)

function render() {
  const walletState = client.wallet.getState()

  // get all registered wallets
  const wallets = walletState.wallets

  // connect to one
  const someWallet = wallets[0];
  client.wallet.connect(someWallet)

  // are we connected?
  if(walletState.connected) {}

  // do we have a signer?
  if(walletState.connected?.signer) {}

  // if we do, send a transaction
  await client.sendTransaction(
    getTransferSolInstruction({
      destination: DEMO_RECIPIENT,
      amount: lamports(10_000_000n), // 0.01 SOL
   }),
);

// transaction fee payer was selected wallet, set to client.payer
// transferSol source was selected wallet, set to client.identity

// can use `walletWithoutSigner` to not set these, or `walletPayer` or `walletIdentity` to just set one
```

## Key functionality

- Access all registered wallets. This filters by chain and `standard:connect` feature, as well as an optional `filter` that can be passed to the plugin. Returns `UiWallet[]`
- After `connect(wallet: UiWallet)`, we derive a `Signer` for the connected wallet account. This will include all available signer features of modify + send tx, sign + send tx, modify + sign message. See Kit function: https://github.com/anza-xyz/kit/blob/f055201c2dd3a4a69b9894d66b622ae81c13b8cd/packages/wallet-account-signer/src/wallet-account-signer.ts#L68. If the wallet account is read-only, we don't set the signer but do set wallet + account
- Four plugins: `walletSigner`, `walletIdentity`, `walletPayer`, `walletWithoutSigner`. All provide an identical `client.wallet` interface. The only difference is which of `client.identity` and `client.payer` they control.
- The connected wallet + account is persisted for autoconnect. By default it persists to localstorage, but anything that matches a subset of [the WebStorage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API) works. Can be disabled by passing `storage: null`, and is always disabled on the server. `autoConnect` can be independently disabled while using storage too.
- State is exposed through the `getState()` and `subscribe()` APIs. `getState` returns a referentially stable object describing the state, and `subscribe` allows registering listeners that are notified whenever state changes. Together these enable a reactive store like React's `useSyncExternalStore`, Svelte's `readable` and Vue's reactive state.
- The plugin is SSR safe. On the server it will always be in a `pending` state, and all functions will throw. But consumers can safely include it in their SSR pass for simplicity. The `pending` state is intentionally distinct from `disconnected` to allow eg not rendering/rendering a skeleton and avoid a UI flash from disconnected -> connected 

## Implementation notes/TODOs

- ~~We haven't published Kit since merging `withCleanup`, so for now this plugin implements `[Symbol.Dispose]` directly.~~

- ~~I think we will need one new error added to Kit. For now this is just included as an error class in the plugin. I think it's justifiable in Kit because it'd be useful to any similar wallet functionality. ~~
```ts
SOLANA_ERROR__WALLET__NOT_CONNECTED
// context: { operation: string }
// message: "Cannot $operation: no wallet connected"
```

- ~~We'd also want to add `ClientWithWallet` to kit plugin-interfaces later, and move some of the types there~~ (I noticed we can't do this because of wallet-standard types, so `ClientWithWallet` will be exported from the plugin)

## More examples

I've had Claude generate a simple example app for React, Svelte, Vue and Vanilla JS: https://github.com/anza-xyz/kit-plugins/tree/wallet-plugin-demos-1/examples/wallet-demo/src . These are intended to show how different frameworks can bind to the state in this plugin and expose it to apps. They are not intended to be representative of how actual apps would be written with kit-plugins, but hopefully show how thin the framework-specific layer for wallets can be. One of my design goals is to never again upset someone who doesn't want to use React!

----

For full transparency, this PR includes a spec file. This is long and the result of lots of back and forth with an LLM, I don't expect anyone to review it and will later remove it from the repo. But wanted to include it here to give more context about how this will all work!


amilz (commented):
Awesome. Great work w/ this!

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1299 @@
...
+The `wallet` plugin and `payer` plugin both set `client.payer`. Use `payer()` for static signers (backend keypairs, scripts) and `wallet()` for dApps with user-facing wallet connections.
+
+When both are needed (e.g. a fallback backend signer with wallet override), `payer` should come first in the plugin chain. The wallet plugin captures whatever is in `client.payer` at installation time as its fallback:
+
+```typescript
+const client = createEmptyClient()
+  .use(rpc('https://api.mainnet-beta.solana.com'))
+  .use(payer(backendKeypair))                     // static payer set first
+  .use(wallet({ chain: 'solana:mainnet' }))       // captures keypair as fallback
+  .use(sendTransactions());
```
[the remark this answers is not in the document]
mcintyre94: Yep you're right, I spotted that in the examples it generated too. Will update in the next rev of the spec!

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1299 @@
...
++---------------------------------------------------+
+```
+
+### Relationship to `@solana/react`
+
+This plugin extracts the wallet management logic currently embedded in `@solana/react`'s `SelectedWalletAccountContextProvider` (persistence, auto-restore, account selection, signer creation) into a framework-agnostic layer. Once this plugin ships, `@solana/react` can be rewritten as a thin adapter over `client.wallet.subscribe` / `client.wallet.getSnapshot` — a handful of hooks rather than a full wallet management implementation. The same approach applies to Vue, Svelte, and Solid adapters, all consuming the same plugin.
+
+### Relationship to `payer`
+
+The `wallet` plugin and `payer` plugin both set `client.payer`. Use `payer()` for static signers (backend keypairs, scripts) and `wallet()` for dApps with user-facing wallet connections.
```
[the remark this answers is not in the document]
mcintyre94: That said... I've just realised that doesn't work correctly because of what we discussed earlier:

```ts
createLocalClient()
.use(wallet())
.use(systemProgram())
```

Now `client.system.instructions.whatever().sendTransaction()` is going to use the transaction planner on `createLocalClient`, which was set before the wallet plugin, and is going to use the payer from `createLocalClient`.

Even though we declare system program after wallet, its dependency came before wallet.

At that point you are needing to unbundle to put the wallet plugin before `rpcTransactionPlanner` and having payer and then wallet is kinda questionable.

You also see this with the non-local rpc client:

```ts
createClient({payer: await generateKeyPairSigner()}) // payer is required but we don't really have one...
.use(wallet())
.use(systemProgram())
```

Now the planner has closed over a randomly generated `client.payer` with no funds as its default.

Maybe we do need to just accept that this is going to need to be used in a `createBrowserClient` or similar, and can't just be added after a different client. At that point we might as well use this *instead* of `payer`, before `rpcTransactionPlanner`. And then there's probably no need for the fallback being on the wallet plugin.

trevor-cortex (commented):
## Summary

This PR scaffolds the `@solana/kit-plugin-wallet` package — a framework-agnostic wallet plugin that manages wallet discovery, connection lifecycle, signer creation, and payer integration via wallet-standard. The implementation is stub-only (`createWalletStore` throws `not implemented`), so this is purely the type system, API surface, configuration design, and package scaffolding.

The design is very well thought out. The API surface is clean, the `subscribe`/`getSnapshot` pattern for framework integration is the right call, and the SSR story with `pending` status is elegant. The spec included in `wallet-plugin-spec.md` provides excellent context for the implementation that will follow.

## Key observations

### Things that look great
- **`hasSigner` in snapshot vs `signer` in `WalletConnection`** — smart separation to avoid unnecessary re-renders while still exposing the actual signer for instruction building.
- **`pending` vs `disconnected` distinction** — avoids the UI flash problem on SSR/auto-reconnect. This is a common pain point in wallet adapters.
- **`usePayer` with fallback capture** — the dynamic getter approach with fallback to the previous `client.payer` is clean. Plugin ordering (`payer` before `wallet`) is well-documented.
- **`WalletStorage` interface** — supporting async backends while being compatible with `localStorage`/`sessionStorage` directly is practical.
- **`signIn` overloads** — the two-argument form (SIWS-as-connect) is a nice ergonomic touch.

### Things to watch during implementation
- **Race conditions in `connect`/`disconnect`** — concurrent calls (user spam-clicks connect, or connects during auto-reconnect) need careful state machine transitions. The spec mentions this but it'll be the trickiest part of `createWalletStore`.
- **Signer caching invalidation** — when a wallet's accounts change externally (via `standard:events`), cached signers for removed accounts need cleanup. The `WalletStoreState` has a single `signer` field, so this should be straightforward, but worth verifying during implementation.
- **`fallbackPayer` capture timing** — the spec shows capturing `client.payer` at plugin install time. If the prior payer is itself a dynamic getter (from another plugin), you'll want to capture the descriptor, not the value, to preserve the getter chain. Or at least document that the fallback is evaluated once.

### Minor items

See inline comments below.

- The `wallet-plugin-spec.md` (1299 lines) is context-only and not intended for permanent inclusion in the repo — author has noted this.
- The `createWalletStore` function is the heart of this plugin and is entirely stubbed. The real review work comes when that's implemented.
- `@solana/kit` is listed as both a `peerDependency` and resolved as a regular dependency in the lockfile — worth checking if this is intentional or if it should only be a peer dep.
- The `extendClient` import from `@solana/kit` is present but unused in the current scaffold (the plugin function implementation is truncated in the diff, but the store is a stub anyway).

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1270 @@
...
+function buildWalletNamespace(store: WalletStore) {
+    return {
+        subscribe: (l: () => void) => store.subscribe(l),
+        getSnapshot: () => store.getSnapshot(),
+        connect: (w: UiWallet) => store.connect(w),
+        disconnect: () => store.disconnect(),
+        selectAccount: (a: UiWalletAccount) => store.selectAccount(a),
+        signMessage: (msg: Uint8Array) => store.signMessage(msg),
+        signIn: (...args: [SolanaSignInInput?] | [UiWallet, SolanaSignInInput?]) => store.signIn(...args),
+    };
```
lorisleiva: I must be missing something but is this not the same as `return store`?

mcintyre94:
I've made another update:

- `getSnapshot` renamed to `getState` as discussed
- the internal store is simplified to only cover the public API, meaning we just return it as `wallet`. The only extra thing it includes is the implementation of `Disposable` which is harmless. We remove `buildWalletNamespace` which mapped from the store to the wallet public API
- `signer` is now only defined in one place: `connected?.signer`, was previously duplicated in the store

mcintyre94:
To flag, I've realised that `ClientWithWallet` cannot be moved to Kit `@solana/plugin-interfaces` as I planned, because of its dependencies on types from `@wallet-standard/ui`. Technically this is all just types and won't affect bundles, but it'd add an NPM dependency that we probably don't want.

For now I think it'll be best to just leave these types in the plugin itself, as the current version of this PR does.

mcintyre94:
I've split the code into three different files (types/store/wallet). This is annoying for the diff, but better to do it here than in the implementation. Note that types and wallet are exported from the package, store is not.

packages/kit-plugin-wallet/src/index.ts
```diff
@@ -0,0 +1,455 @@
...
+            signIn: store.signIn,
+            signMessage: (msg: Uint8Array) => store.signMessage(msg),
+            subscribe: (l: () => void) => store.subscribe(l),
+        };
+
+        // Define getters for the rest of the state properties
+        for (const [key, fn] of Object.entries({
+            connected: () => store.getConnected(),
+            status: () => store.getState().status,
+            wallets: () => store.getState().wallets,
```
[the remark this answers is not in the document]
mcintyre94: Resolving - all state is now in `getState` only.

packages/kit-plugin-wallet/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "scripts": {
+        "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
+        "dev": "vitest --project node",
+        "lint": "eslint . && prettier --check .",
+        "lint:fix": "eslint --fix . && prettier --write .",
+        "test": "pnpm test:types && pnpm test:treeshakability",
+        "test:treeshakability": "for file in dist/index.*.mjs; do agadoo $file; done",
+        "test:types": "tsc --noEmit"
+    },
+    "peerDependencies": {
```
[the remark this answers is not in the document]
mcintyre94: Resolving - we use this peer dependency for all plugins

packages/kit-plugin-wallet/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://json.schemastore.org/tsconfig",
+    "compilerOptions": { "lib": [] },
```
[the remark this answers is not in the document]
mcintyre94: Resolving - matches current plugins. We'll update as needed when implementing

wallet-plugin-spec.md
mcintyre94: Resolving - store now matches the public API and this awkward mapping is gone. Keeping https://github.com/anza-xyz/kit-plugins/pull/173#discussion_r3029757073 open for `client.identity`.

trevor-cortex (commented):
## Summary

This PR scaffolds the `@solana/kit-plugin-wallet` package, establishing the type system, plugin entry points (`wallet` and `walletAsPayer`), and a stub store — without yet implementing the wallet discovery/connection logic. The design is well thought out: clean separation between types, store, and plugin wiring; a `subscribe`/`getState` API that maps directly onto framework primitives (`useSyncExternalStore`, Svelte `readable`, Vue `shallowRef`); SSR safety via the `pending` state; and a pluggable storage adapter for persistence.

## Key observations

**Design looks solid.** The split into `wallet()` vs `walletAsPayer()` is clean and gives users the right level of control. The `WalletSigner` type alias is a nice addition. The `Object.defineProperty` approach for the dynamic payer getter in `walletAsPayer` is the right call.

**Types are thorough and well-documented.** The JSDoc throughout `types.ts` is excellent — each type, field, and method has clear documentation with examples where helpful.

## Things to watch for in subsequent reviews / during implementation

- The `WalletNamespace` JSDoc references `{@link getSnapshot}` (line 137 of types.ts) but the method is called `getState`. Should be `{@link getState}`.
- The `payer` getter in `walletAsPayer` coerces `null` signer to `undefined` via `?? undefined`. This is correct for the type (`TransactionSigner | undefined`) but worth a comment explaining why — `connected?.signer` can be `null` for read-only wallets, while the `payer` slot expects `undefined` for "no payer".
- `@solana/kit` appears as a `peerDependency` but also gets resolved as a regular dependency in the lockfile (the `packages/kit-plugin-wallet` section in `pnpm-lock.yaml` lists it under `dependencies`). This seems to be a pnpm workspace resolution artifact, but worth double-checking it doesn't cause duplicate bundling.
- The spec file (`wallet-plugin-spec.md`) is at the repo root. The PR description mentions it'll be removed later — might be worth adding a TODO comment or tracking issue so it doesn't get forgotten.
- `tsconfig.json` uses `lib: []` which is consistent with the existing `kit-plugins` package. The implementation will need DOM types (for `localStorage`, `window` checks, etc.) — this will presumably be handled via the `__BROWSER__` global declarations and conditional compilation through tsup, matching the existing pattern.

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -0,0 +1,297 @@
...
+     * loads. Pass `null` to disable persistence entirely.
+     *
+     * When omitted in a browser environment, `localStorage` is used by default.
+     * On the server, storage is always skipped regardless of this option.
+     *
+     * @default localStorage (in browser)
+     * @see {@link WalletStorage}
+     */
+    storage?: WalletStorage | null;
+
```
trevor-cortex, line 176:
```suggestion
 * All wallet state is accessed via {@link getState}. Use {@link subscribe}
```

The `@link` references `getSnapshot` but the method is named `getState`.

packages/kit-plugin-wallet/src/wallet.ts
```diff
@@ -0,0 +1,134 @@
...
+        const obj = withCleanup(
+            extendClient(client, {
+                wallet: store,
+            }),
+            () => store[Symbol.dispose](),
+        );
+
+        Object.defineProperty(obj, 'payer', {
+            configurable: true,
+            enumerable: true,
```
trevor-cortex: Might be worth a brief comment here explaining the `?? undefined` — `connected?.signer` can be `null` (read-only wallet) but the `payer` type expects `undefined` for "no payer available".

```suggestion
                return obj.wallet.getState().connected?.signer ?? undefined; // null signer → undefined payer
```

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * Thrown when a wallet operation is attempted but no wallet is connected
+ * (or the connected wallet is read-only and has no signer).
+ *
+ * @example
+ * ```ts
+ * try {
+ *   await client.wallet.signMessage(message);
+ * } catch (e) {
+ *   if (e instanceof WalletNotConnectedError) {
```
trevor-cortex: nit: Since this error class may eventually move to Kit proper (per the TODO on line 269), consider adding a static discriminant that survives minification:

```ts
static readonly code = 'WALLET_NOT_CONNECTED' as const;
readonly code = WalletNotConnectedError.code;
```

This would let consumers check `error.code === 'WALLET_NOT_CONNECTED'` instead of relying on `instanceof`, which can break across package versions or bundler deduplication. Not critical for the scaffold, but good to keep in mind for the implementation pass.

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -0,0 +1,297 @@
...
+import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';
+import type { UiWallet, UiWalletAccount } from '@wallet-standard/ui';
+
+/**
+ * The signer type for a connected wallet account.
+ *
+ * Always satisfies `TransactionSigner`. Additionally implements `MessageSigner`
+ * when the wallet supports `solana:signMessage`.
+ */
+export type WalletSigner = TransactionSigner | (MessageSigner & TransactionSigner);
```
trevor-cortex, line 13: 👍 Nice — extracting this as a named type alias cleans up the repeated union in `WalletState.connected.signer` and the spec.

mcintyre94:
* **#173** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/173?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
Added another commit: cb6f652308dd605836e02b8eeb9992ad47d1a285 

Addresses issues found in the spec/scaffold while implementing the plugin:

- `signMessage` now uses the feature directly, instead of using the cached `connected.signer`. Academically this means that a wallet that has the signMessage feature but no transaction feature, works correctly. But more practically this simplifies the implementation and removes an error case that doesn't cleanly fit the existing errors
-  SIWS is simplified to require passing `UiWallet`, the same as connect. Previously we had a version that automatically signs in an account on the connected wallet. This doesn't really need a special case (you can just pass `connected.wallet`) and removing it simplifies the code a lot. I think SIWS as connect is the majority use case too.
- Fixes a bug where we try to extend a frozen object from `extendClient`

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+```tsx
+import { useSyncExternalStore } from 'react';
+
+function useWalletState() {
+    return useSyncExternalStore(client.wallet.subscribe, client.wallet.getState);
+}
+
+function App() {
+    const { wallets, connected, status } = useWalletState();
```
trevor-cortex, line 174: The README shows `signIn(input?)` and `signIn(wallet, input?)` as two separate overloaded forms, but `WalletNamespace.signIn` in `types.ts` only has the two-argument signature `(wallet: UiWallet, input: SolanaSignInInput)`. The spec also only defines the two-argument form and explicitly calls out the design choice to avoid overloads.

Should the README be updated to match the types? e.g.:
```ts
// Sign in and connect in one step
const output = await client.wallet.signIn(selectedWallet, { domain: window.location.host });

// Sign in with the already-connected wallet
const output = await client.wallet.signIn(client.wallet.getState().connected.wallet, { domain: window.location.host });
```

trevor-cortex (commented):
## Re-review — scaffold looks good

This is a re-review after changes since my last pass. The previous issues I raised (stale `getSnapshot` JSDoc link, missing comment on the `?? undefined` coercion) have both been addressed. The code is in good shape for a scaffold PR.

### What changed
- File structure split into `types.ts`, `wallet.ts`, `store.ts` — much cleaner than the single-file version.
- The `wallet.ts` plugin functions now properly use `withCleanup` wrapping `extendClient` rather than spread.
- `store.ts` is a clean stub with the right shape (`WalletStore = WalletNamespace & { [Symbol.dispose] }`).

### New observations
- **README `signIn` docs vs types mismatch**: The README shows two forms — `signIn(input?)` (connected wallet) and `signIn(wallet, input?)` (connect-and-sign-in). But `WalletNamespace.signIn` in `types.ts` only has the two-argument form `(wallet: UiWallet, input: SolanaSignInInput)`. The README implies an overload that doesn't exist in the types. The spec also only has the two-argument form and explains the design choice. I'd update the README to match.
- **`store.ts` not exported from `index.ts`** — intentional and correct (marked `@internal`), just confirming I noticed.

### Previously raised — still relevant
- **Error class resilience**: `WalletNotConnectedError` still relies on `this.name = 'WalletNotConnectedError'` which can break under minification if someone does `error.name === 'WalletNotConnectedError'`. A static `code` property (like Kit's `SolanaError`) would make this robust. The TODO comment about moving this to Kit proper addresses this long-term.
- **`@solana/kit` in lockfile**: Still appears as a resolved regular dep under `packages/kit-plugin-wallet` in the lockfile despite being listed only as a `peerDependency` in `package.json`. Likely a pnpm workspace artifact, but worth verifying it doesn't cause duplicate bundling.

### For subsequent reviewers
- The spec file at the repo root (`wallet-plugin-spec.md`) is intended to be temporary — the PR description mentions removing it later.
- The `tsconfig.json` uses `lib: []` — this is consistent with the repo pattern; DOM types come in via the `__BROWSER__` compile-time constants and tsup's multi-target build.
- The store implementation is entirely stubbed — the interesting review will be when `createWalletStore` gets its real implementation.

packages/kit-plugin-wallet/README.md
mcintyre94: Spotted this too, fixed in upstack PR

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1237 @@
...
+}
+```
+
+The bridge function (`createSignerFromWalletAccount`) is consumed via `@solana/wallet-account-signer`. `extendClient` and `withCleanup` are consumed via `@solana/kit`.
+
+## Summary
+
+A framework-agnostic Kit plugin that manages wallet discovery, connection lifecycle, and signer creation using wallet-standard. When a wallet is connected, the plugin optionally syncs its signer to the client's `payer` slot via a dynamic getter, falling back to any previously configured payer when no wallet is connected. The plugin exposes subscribable wallet state for framework adapters (React, Vue, Svelte, etc.) to consume without coupling to any specific UI framework.
+
+**SSR-safe.** Both `wallet` and `walletAsPayer` can be included in the same client chain on both server and browser. On the server (`typeof window === 'undefined'`), the plugin gracefully degrades — status stays `'pending'`, wallet list is empty, payer is `undefined` (when using `walletAsPayer`), storage is skipped, no registry listeners are created. On the browser it initializes fully. This means a single client chain works everywhere:
```
lorisleiva: I am unsure about this design as we end up with a type-safe plugin that's lying about its type and causes runtime failures.

It could be tricky to implement but we could instead introduce a plugin wrapper that splits two plugin based on its environment if and only if they return the same `Client` output (so the rest of the plugin chain is type-safe). Something like:

```ts
const client = createClient().
    .use(ssr({
        client: walletSigner({ chain: 'solana:mainnet' })
        server: signerFromFile("~/.solana/config/id.json")
    }))
    .use(solanaRpc('https://api.mainnet-beta.solana.com'))
    .use(systemProgram());
```

Alternatively, we'd expect the server-side to instantiate a different Kit client than the client-side which I think is a fair assumption and makes since less complex.

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1237 @@
...
+    // (not defined after the fact) because extendClient freezes the result.
+    const additions = {
+      wallet: store,
+      ...withCleanup(client, () => store.destroy()),
+    };
+
+    Object.defineProperty(additions, 'payer', {
+      get() {
+        return store.getState().connected?.signer ?? undefined;
+      },
```
lorisleiva: I don't think a plugin that adds `Partial<ClientWithPayer>` to the client is going to be very useful because none of the plugin requiring `ClientWithPayer` will accept it.

I think it's best to return a `ClientWithPayer` directly and, when the wallet isn't connected, throw an error when trying to access `client.payer` with a message showing the wallet status.

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1237 @@
...
+
+// Wallet alongside a static payer
+const client = createEmptyClient()
+  .use(rpc('https://api.mainnet-beta.solana.com'))
+  .use(payer(backendKeypair))
+  .use(wallet({ chain: 'solana:mainnet' }))
+  .use(planAndSendTransactions());
+// client.payer is TransactionSigner (from payer plugin, untouched)
+// client.wallet.getState().connected?.signer for manual use
+```
```
lorisleiva: As discussed offline, [the new `@solana/kit-plugin-signer` package](https://github.com/anza-xyz/kit-plugins/pull/189) provides a family of helpers for each kind of signer strategy. For instance, `generatedPayer`/`generatedIdentity`/`generatedSigner` all add a new generated signer to the client such that the first one sets `client.payer`, the second sets `client.identity` and the last sets both using the same signer object (the one most users will end up using).

As such, we should aim to unify these APIs since the wallet plugin is able to set signers on the client too.

The main difference is the wallet plugin also sets `client.wallet` which acts as a "control centre" for managing connected wallets. Therefore we could go with something like:

- `walletSigner`: Sets `client.wallet`, `client.payer` and `client.identity` (the most popular entrypoint).
- `walletIdentity`: Sets `client.wallet`, `client.identity` (the second most popular entrypoint for dApps that have a different payer signer).
- `walletPayer`: Sets `client.wallet`, `client.payer`.
- `walletWithoutSigner`: Sets `client.wallet` only (more verbose to discourage users to default to using it).

One other thing to consider is: **could we technically select two different wallets as the `client.identity` and `client.payer` signers?** If so, we should consider if it'd make sense for `client.identity` and `client.payer` to own their own `client.wallet` stores internally (or a subset of it).

wallet-plugin-spec.md
```diff
@@ -0,0 +1,1237 @@
...
+| `signMessage` on wallet without feature | `getWalletFeature` throws `WalletStandardError` |
+| `signIn` on wallet without feature | `getWalletFeature` throws `WalletStandardError` |
+
+`connect()` and `disconnect()` propagate wallet errors to the caller unchanged. Internal errors (reconnect failures, storage errors) are logged via `console.warn` but do not throw.
+
+## Design Decisions
+
+**Descriptor-preserving composition.** The plugin uses `extendClient` from plugin-core to build the client, preserving getters and symbol-keyed properties from previous plugins. The updated `addUse` also preserves descriptors through `.use()` calls. Downstream plugins should use `extendClient` rather than spread to avoid flattening the dynamic payer getter.
+
+**Single chain per client.** Signers are bound to a specific chain at creation time. Switching chains requires a different RPC endpoint too. One client = one network.
```
lorisleiva, line 1290: See for me if we have the design principle "1 client = 1 network", then I don't think it's unexpected to also have the design principle "1 client = 1 environment" (i.e. client vs server).

packages/kit-plugin-wallet/src/index.ts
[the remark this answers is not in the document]
mcintyre94: Addressed in upstack PR

wallet-plugin-spec.md
mcintyre94: The use case I'm really trying to support here is SSR hydration, not actually using the client with a wallet in both a server and browser environment. I want you to be able to render the page on the server with the pending state, and for that not to have hydration mismatches when it then renders on the browser. And then actually identifying and connecting to wallets if they are available in the browser is progressive UI/capabilities.

If you actually want to use the client from a server environment (eg API routes or framework server actions), then I agree that should be a separate client that wouldn't use wallet at all.

I'm just trying to avoid forcing people to use client-only components and syncing a separate server placeholder, rather than just rendering the same client and pending state for the same page in both SSR and the browser.

wallet-plugin-spec.md
mcintyre94: I'm happy with doing it that way too. I didn't for the same concern you mentioned above - claiming to implement the type but then throwing a runtime error.

I'm not sure if this changes in your open PRs but I based this decision on the fact that plugins like `rpcTransactionPlanner` were using `Partial<ClientWithPayer>` as their input client type, so it would be accepted for that.

I don't feel strongly either way though, especially if that has changed/will change.

wallet-plugin-spec.md
mcintyre94: Agree that if we add all these entrypoints then we should make  the `walletWithoutSigner` more verbose.

I'd lean against exposing different `client.wallet` stores for identity and payer though. Technically they could be independent signers, both implemented by a wallet. But that doesn't require different `client.wallet` stores, the `Signer` contains everything they need, the fact they internally call to a wallet is implementation detail.

While the `client.wallet` store only stores one selected wallet/account/signer in `client.wallet.connected`, you have access to all connected accounts for all wallets in `client.wallet.getState().wallets`. And you can use `createSignerFromWalletAccount` from Kit to get a signer for any of those.

So I think in the unusual case that you want to have signers for multiple wallet accounts and set them as payer/identity, you have everything you need with just one `client.wallet` store, and your own custom identity/payer logic.

The one thing that would be arguably useful is to allow them to store the equivalent of `client.wallet.connected`, ie the wallet/account/signer. But I'm not sure how we could structure payer/identity to handle that without complicating all the non-wallet use cases, so I think app specific logic is probably the right way to deal with that.

wallet-plugin-spec.md
lorisleiva: That `rpcTransactionPlanner` design is more of an exception than the rule and to be honest, now that we're fully leaning into the plugin design (i.e. not using client factories), I should totally change that plugin requirement so it requests `ClientWithPayer` instead of throwing a runtime error when both are not provided.

Just about every other plugin (including the program plugins) will end up requesting `ClientWithPayer` so `Partial<ClientWithPayer>` isn't gonna take us very far.

I think if the goal is making sure SSR hydration works but really you're designing the client plugin, then I think we'll make the developer experience more enjoyable with `ClientWithPayer`.

wallet-plugin-spec.md
lorisleiva: I agree that having a single `client.wallet` store makes sense even if you have multiple selected wallets as `client.payer` and `client.identity`. It's just the `connected` state I'm unsure of (_if_ we end up supporting multiple wallet selected). Theoretically there's more information you could store inside a `client.payer` and `client.identity` so that could move there. I'm just not sure what the high level API looks like for connecting two different wallets. Is that two plugins: `walletPayer` and `walletIdentity` and the second call doesn't override the existing `client.wallet`? How do you then select a payer wallet versus an identity wallet?

Maybe this is out-of-scope for now though and, if this is needed in the future, it would materialise as a separate plugin. I just wanted to make sure we were not missing a design piece by not analysing this use-case.

wallet-plugin-spec.md
lorisleiva: Gotcha! I'll resolve this since it's related to our other conversation below.

wallet-plugin-spec.md
mcintyre94: In practice an app juggling multiple connected wallets would probably want to just use wallet-standard directly:

```ts
const connectFeature = getWalletFeature(
    uiWallet,
    StandardConnect,
) as StandardConnectFeature[typeof StandardConnect];

// Snapshot existing accounts before connect.
const existingAccounts = [...uiWallet.accounts];

const updatedAccounts = await connectFeature.connect();

// persist connected accounts / convert to signer if needed
```

So you'd bypass `client.wallet.getState().connected`, and just be using the wallet plugin for tracking registered wallets/all their connected accounts/event handlers etc. and handle mapping them to signers yourself.

There is a missing piece there for setting those signers to identity/payer though, but I think that being out of scope is ok. If you wanted to do it yourself you'd probably create a custom plugin that exposes a setter for each of them, and then in your app logic where you handle connecting wallets you'd call those setters.

wallet-plugin-spec.md
mcintyre94: Makes sense, I'm happy to update to set `ClientWithPayer` (and also `ClientWithIdentity` when we add that)

lorisleiva: Made a quick PR for the `Partial<ClientWithPayer>` removal: https://github.com/anza-xyz/kit-plugins/pull/192.

mcintyre94: Updated this, it now returns `ClientWithPayer`. I'll open a PR on Kit with the errors proposed for the case where it isn't set.

wallet-plugin-spec.md
mcintyre94: Updated to the suggestion at the top of this thread:

- walletSigner: Sets client.wallet, client.payer and client.identity (the most popular entrypoint).
- walletIdentity: Sets client.wallet, client.identity (the second most popular entrypoint for dApps that have a different payer signer).
- walletPayer: Sets client.wallet, client.payer.
- walletWithoutSigner: Sets client.wallet only (more verbose to discourage users to default to using it).

wallet-plugin-spec.md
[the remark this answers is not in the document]
mcintyre94: Resolving this one - we now have `walletIdentity` and `walletSigner`.

packages/kit-plugin-wallet/src/types.ts
mcintyre94: Resolving - error moved to Kit

wallet-plugin-spec.md
mcintyre94: This would be the same as https://github.com/anza-xyz/kit-plugins/pull/173#discussion_r3080100249 - treating SSR as an environment where you should be able to use the same client as the browser for developer convenience.

packages/kit-plugin-wallet/src/wallet.ts
```diff
@@ -0,0 +1,179 @@
...
+        },
+    });
+}
+
+function createPlugin<TAdditions extends ClientWithWallet>(config: WalletPluginConfig, signerProperties: string[]) {
+    return <T extends object & { wallet?: never }>(
+        client: T,
+    ): Disposable & Omit<T, 'identity' | 'payer' | 'wallet'> & TAdditions => {
+        if ('wallet' in client) {
+            throw new Error(
```
trevor-cortex, line 34: The `Omit<T, 'identity' | 'payer' | 'wallet'>` always strips both `identity` and `payer` from the incoming type, even for plugins that don't set them. This means:

```ts
const client = createClient()
  .use(identity(myKeypair))           // sets client.identity
  .use(walletWithoutSigner({ ... })); // should only add wallet

client.identity; // ← type error, but the value is still there at runtime
```

Same issue with `walletPayer` stripping `identity` and `walletIdentity` stripping `payer` from the preceding chain.

The `Omit` should only strip properties that the plugin actually re-adds. One approach:

```ts
function createPlugin<TAdditions extends ClientWithWallet>(
  config: WalletPluginConfig,
  signerProperties: string[],
) {
  return <T extends object & { wallet?: never }>(
    client: T,
  ): Disposable & Omit<T, keyof TAdditions> & TAdditions => { ... };
}
```

That way `walletWithoutSigner` only omits `wallet`, `walletPayer` omits `wallet` + `payer`, etc.

trevor-cortex (commented):
## Re-review — Scaffold/design for the wallet plugin

### What changed since last review

Big design evolution: the two plugins (`wallet`, `walletAsPayer`) are now four (`walletSigner`, `walletPayer`, `walletIdentity`, `walletWithoutSigner`). This is a much cleaner design — the 4-way matrix covering `payer` × `identity` gives users exactly the right level of control without needing to understand the internals.

The `createPlugin` + `defineSignerGetter` pattern in `wallet.ts` is elegant — one generic factory with the signer properties parameterized. The two distinct error cases (no wallet connected vs. read-only wallet) in the getter are a nice touch.

### Issues

**🐛 Type bug: `Omit<T, 'identity' | 'payer' | 'wallet'>` is over-broad** — see inline comment. This will strip `payer`/`identity` from the input type even for plugins that don't replace them.

**📝 Stale JSDoc references** — a couple of `{@link wallet}` / `{@link walletAsPayer}` references survived the rename. See inline comments.

### Still relevant from prior reviews

- `@solana/kit` appearing as a resolved regular dependency in the lockfile despite being declared as `peerDependencies` only in `package.json`. The `packages/kit-plugin-wallet` section in `pnpm-lock.yaml` lists it under `dependencies`. This is likely a pnpm workspace resolution artifact, but worth verifying it doesn't cause duplicate bundling for consumers.

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -0,0 +1,248 @@
...
+    setItem(key: string, value: string): Promise<void> | void;
+};
+
+/**
+ * Configuration for the {@link wallet} and {@link walletAsPayer} plugins.
+ */
+export type WalletPluginConfig = {
+    /**
+     * Whether to attempt silent reconnection on startup using the persisted
+     * wallet account from `storage`.
```
trevor-cortex, line 124: Stale reference: `{@link wallet}` and `{@link walletAsPayer}` no longer exist. Should reference the new plugin names (`walletSigner`, `walletPayer`, `walletIdentity`, `walletWithoutSigner`).

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -0,0 +1,248 @@
...
+    subscribe: (listener: () => void) => () => void;
+};
+
+/**
+ * Properties added to the client by the {@link wallet} plugin.
+ *
+ * All wallet state and actions are namespaced under `client.wallet`.
+ * `client.payer` is not affected — use the {@link walletAsPayer} plugin to
+ * set the payer dynamically from the connected wallet.
+ *
```
trevor-cortex, line 292: Stale JSDoc — references `{@link wallet}` and `{@link walletAsPayer}`, should be updated to the current names (e.g. `{@link walletPayer}` or `{@link walletSigner}`).

packages/kit-plugin-wallet/src/wallet.ts
mcintyre94: Fixed + added typetests to verify the behaviour of types of these plugins (🤞 these won't summon copilot spam)

trevor-cortex (commented):
## Re-review — Scaffold/design for the wallet plugin

Both issues from my last review have been addressed:

- **`Omit` type bug** — Fixed. `createPlugin` now correctly uses `Omit<T, keyof TAdditions>`, so each variant only strips the properties it actually adds. ✅
- **Stale JSDoc references** — Fixed. `WalletPluginConfig` docs now correctly reference `walletSigner`, `walletPayer`, `walletIdentity`, `walletWithoutSigner`. ✅

### Type tests

Great addition. The new `wallet-typetest.ts` covers all four plugin variants and validates:
- Each plugin adds the expected properties (`payer`, `identity`, `wallet`)
- Plugins don't strip previously-set properties they don't own (the `Omit` fix)
- Only one wallet plugin can be used per client (via `{ wallet?: never }` constraint)

The scaffold looks good to me — no new issues found.

### Still noting (non-blocking)

- `@solana/kit` appearing as a resolved regular dependency in `pnpm-lock.yaml` despite being declared peer-only in `package.json`. Likely a pnpm workspace artifact, but worth a quick check that consumers don't get a duplicate bundle.

lorisleiva (approved):
Nice, thanks for making the `walletIdentity`/`walletPayer`/`walletSigner` changes to be consistent with `@solana/kit-plugin-signers`.

mcintyre94:
Pushed a small change to widen the type of `chain` to `SolanaChain | IdentifierString & {}`. Ie you can pass any `IdentifierString`, matching wallet-standard, and `SolanaChain` ones will still autocomplete in IDEs. For now we'll need to cast to `SolanaChain` when we pass it to Kit, but that's just a type issue too. I have a PR open to widen the type equivalently in Kit, but we can land that independently. https://github.com/anza-xyz/kit/pull/1548

PR #184, opened 2026-04-12
Add granular RPC connection plugins and deprecate `rpc`/`localhostRpc`

lorisleiva:
This PR introduces four new granular plugins for setting up RPC and RPC Subscriptions connections independently: `rpcConnection`, `rpcSubscriptionsConnection`, `solanaRpcConnection`, and `solanaRpcSubscriptionsConnection`. The existing `rpc` and `localhostRpc` plugins, which inject both RPC and RPC Subscriptions in one go, are deprecated in favor of the new alternatives. Docblocks with migration snippets are included on the deprecated functions.

PR #183, opened 2026-04-12
Bump `@solana/kit` peer dependency to ^6.8.0

lorisleiva:
This PR bumps the `@solana/kit` peer dependency to ^6.8.0 across all packages and updates all references to the renamed `createEmptyClient` API in source files, tests, docblocks, and READMEs. No behavioral changes.

PR #185, opened 2026-04-12
Add all-in-one Solana RPC plugins

lorisleiva:
This PR adds `solanaRpc`, `solanaMainnetRpc`, `solanaDevnetRpc`, and `solanaLocalRpc` — all-in-one plugins that set up a full Solana RPC client with RPC connection, subscriptions, minimum balance computation, transaction planning, and transaction execution in a single call. Cluster-specific presets provide sensible defaults and the devnet/local variants include airdrop support.


lorisleiva:
* **#190** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/190?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#189** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#188** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#187** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/187?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#186** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#185** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/185?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/185?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#184** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#183** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/183?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#173](https://github.com/anza-xyz/kit-plugins/pull/173) <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/173?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/kit-plugin-rpc/src/solana-rpc.ts
```diff
@@ -1,12 +1,215 @@
...
+ * @see {@link solanaLocalRpc}
+ */
+export function solanaRpc<TClusterUrl extends ClusterUrl>(config: SolanaRpcConfig<TClusterUrl>) {
+    return <T extends ClientWithPayer>(client: T) =>
+        pipe(
+            client,
+            solanaRpcConnection<TClusterUrl>(config.rpcUrl, config.rpcConfig),
+            solanaRpcSubscriptionsConnection<TClusterUrl>(
+                config.rpcSubscriptionsUrl ?? (config.rpcUrl.replace(/^http/, 'ws') as TClusterUrl),
+                config.rpcSubscriptionsConfig,
```
trevor-cortex, line 96: Nit: the regex `/^http/` will also match `https` → `wss` which happens to be correct, but it's slightly implicit. If someone passes a URL like `httpbin.org/...` it would produce `wsbin.org/...`. Probably fine for the intended use case, but worth noting.

packages/kit-plugin-rpc/src/solana-rpc.ts
```diff
@@ -1,12 +1,215 @@
...
+ * import { solanaMainnetRpc } from '@solana/kit-plugin-rpc';
+ * import { payer } from '@solana/kit-plugin-payer';
+ *
+ * const client = createClient()
+ *     .use(payer(myPayer))
+ *     .use(solanaMainnetRpc({ rpcUrl: 'https://api.mainnet-beta.solana.com' }));
+ * ```
+ *
+ * @see {@link solanaRpc}
+ * @see {@link solanaDevnetRpc}
```
trevor-cortex, line 127: Minor: `SolanaRpcConfig<string>` here means TypeScript won't enforce that the `rpcUrl` is actually a mainnet URL at the call site — you could pass a devnet URL and it'd happily cast it. This is presumably intentional for flexibility (users with custom RPC providers), but something to be aware of for docs.

trevor-cortex (commented):
## Summary

This PR adds four all-in-one plugins (`solanaRpc`, `solanaMainnetRpc`, `solanaDevnetRpc`, `solanaLocalRpc`) that compose existing lower-level plugins into a single `.use()` call, giving users a complete Solana RPC client with connection, subscriptions, minimum balance, transaction planning, and execution. Cluster-specific presets provide sensible defaults, and devnet/local variants include airdrop support.

## Notes

Clean implementation — the composition via `pipe` reusing existing plugins is well-structured. The new `@solana/kit-plugin-instruction-plan` workspace dependency makes sense for `planAndSendTransactions`.

Looks good to me — a couple of minor observations below, but nothing blocking.

## For subsequent reviewers

- The `solanaMainnetRpc` function accepts `SolanaRpcConfig<string>` and casts to `SolanaRpcConfig<MainnetUrl>` internally — worth verifying this gives the expected type-narrowing downstream.
- The `DefaultRpcSubscriptionsChannelConfig` import is now unused in the new code section (it's still used by the deprecated `rpc` function below, so it's fine for now).
- Tests are structural only (checking properties exist) — no integration tests hitting real RPCs, which is reasonable for this layer.

packages/kit-plugin-rpc/src/solana-rpc.ts
lorisleiva: That's fine.

packages/kit-plugin-rpc/src/solana-rpc.ts
lorisleiva: That's on purpose because otherwise we force the user to provide `mainnet("...")` even though the user already flagged it's on mainnet via the `solanaMainnetRpc` call.

packages/kit-plugin-rpc/src/solana-rpc.ts
```diff
@@ -1,12 +1,215 @@
...
+ * @see {@link solanaMainnetRpc}
+ * @see {@link solanaDevnetRpc}
+ * @see {@link solanaLocalRpc}
+ */
+export function solanaRpc<TClusterUrl extends ClusterUrl>(config: SolanaRpcConfig<TClusterUrl>) {
+    return <T extends ClientWithPayer>(client: T) =>
+        pipe(
+            client,
+            solanaRpcConnection<TClusterUrl>(config.rpcUrl, config.rpcConfig),
+            solanaRpcSubscriptionsConnection<TClusterUrl>(
```
mcintyre94, line 94: I didn't realise you could use `pipe` to compose plugins like this, that's really nice!

mcintyre94 (approved):
Nice!

packages/kit-plugin-rpc/src/solana-rpc.ts
lorisleiva: Yeah `createClient` is just a fancy pipe. In fact I tried using `createClient` first here but it made the types way more complicated than they need to be (because of the async conditionals) so I thought it'd make more sense to stick to pipes for internal plugin compositions.

lorisleiva:
### Merge activity

* **Apr 14, 5:32 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/185).
* **Apr 14, 5:35 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/185) rebased this pull request as part of a merge.
* **Apr 14, 5:36 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/185).

PR #191, opened 2026-04-13
Implement kit-plugin-wallet store, types, and tests

mcintyre94:
Implements the wallet plugin's core store (createWalletStore) with wallet discovery, connection lifecycle, signer creation, and subscribable state management.

Tests are split into store tests (state management, lifecycle) and wallet tests (plugin integration, payer getter).

Also add to list of plugins in root readme and contributing


mcintyre94:
* **#191** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,577 @@
...
+                    currentWallet,
+                    StandardDisconnect,
+                ) as StandardDisconnectFeature[typeof StandardDisconnect];
+                await disconnectFeature.disconnect();
+            }
+        } finally {
+            disconnectLocally();
+        }
+    }
+
```
trevor-cortex, line 492: **Bug: missing `status: 'connecting'` and error handling in `signIn`.**

Unlike `connect()` which sets `status: 'connecting'` and has a try/catch that calls `disconnectLocally()` on failure, `signIn` does neither:

1. UI won't show a loading state during the async `signIn` call
2. If `signInFeature.signIn(input)` throws (user rejects), the store status is never reset — it stays at whatever it was before (`'pending'`, `'disconnected'`, etc.) rather than being cleaned up

Suggestion:
```ts
async function signIn(uiWallet: UiWallet, input: SolanaSignInInput): Promise<SolanaSignInOutput> {
    userHasSelected = true;
    cancelReconnect();
    setState({ status: 'connecting' });
    try {
        const signInFeature = getWalletFeature(...);
        const [result] = await signInFeature.signIn(input);
        // ... rest of success path
    } catch (error) {
        disconnectLocally();
        throw error;
    }
}
```

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,577 @@
...
+    function setState(updates: Partial<WalletStoreState>): void {
+        const prev = state;
+        state = { ...state, ...updates };
+
+        // Only create a new snapshot when a snapshot-relevant field changed.
+        // This ensures referential stability for useSyncExternalStore.
+        if (
+            state.wallets !== prev.wallets ||
+            state.connectedWallet !== prev.connectedWallet ||
+            state.account !== prev.account ||
```
trevor-cortex, line 164: Nit: `listeners.forEach(l => l())` runs unconditionally, even when the snapshot didn't change (i.e. the `if` condition above was false). This means subscribers are notified on every `setState` call even if no snapshot-relevant field changed.

For React's `useSyncExternalStore` this is benign (it does its own `Object.is` check), but other frameworks or vanilla subscribers that re-render on notification will get unnecessary updates. Moving the `listeners.forEach` inside the `if` block would be more correct:

```ts
if (/* snapshot changed */) {
    snapshot = deriveSnapshot(state);
    listeners.forEach(l => l());
}
```

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,577 @@
...
+            }
+
+            const walletName = savedKey.slice(0, separatorIndex);
+            const savedAddress = savedKey.slice(separatorIndex + 1);
+            const existing = state.wallets.find(w => w.name === walletName);
+
+            if (existing) {
+                await attemptSilentReconnect(savedAddress, existing);
+            } else if (
+                registry.get().some(w => {
```
trevor-cortex, line 705: The async IIFE is wrapped in `void`, so if `attemptSilentReconnect` throws an unexpected error (not the ones it catches internally), it'll be silently swallowed as an unhandled rejection. 

Also, if multiple wallets register in rapid succession, this handler fires for each one and multiple concurrent `attemptSilentReconnect` calls can race. Consider adding a guard flag (e.g. `reconnectInProgress`) to prevent concurrent attempts.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,577 @@
...
+        return {};
+    }
+
+    // -- Connection lifecycle ----------------------------------------------
+
+    async function connect(uiWallet: UiWallet): Promise<readonly UiWalletAccount[]> {
+        userHasSelected = true;
+        cancelReconnect();
+        setState({ status: 'connecting' });
+
```
trevor-cortex, line 390: Minor: no guard against concurrent `connect()` calls. If a user clicks two different wallet buttons rapidly:

1. `connect(walletA)` sets status → connecting, starts await
2. `connect(walletB)` sets status → connecting, starts await  
3. walletA resolves → `setConnected(walletA)` sets up `walletEventsCleanup`
4. walletB resolves → `setConnected(walletB)` overwrites `walletEventsCleanup`, leaking walletA's listener

A generation counter or AbortController pattern would prevent stale connect calls from completing. Not a blocker for this PR, but worth tracking.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,577 @@
...
+        const eventsFeature = getWalletFeature(
+            uiWallet,
+            StandardEvents,
+        ) as StandardEventsFeature[typeof StandardEvents];
+
+        return eventsFeature.on('change', properties => {
+            const refreshed = refreshUiWallet(uiWallet);
+
+            // If the wallet no longer passes the filter, disconnect.
```
trevor-cortex: Tiny nit: the `if (currentWallet && ...)` guard is redundant — you already early-return at line 233 if `!state.connectedWallet`, and `currentWallet` is assigned from `state.connectedWallet` on line 235, so it's always truthy here.

trevor-cortex (commented):
## Summary

This PR implements the wallet plugin's core store (`createWalletStore`) with wallet discovery via wallet-standard, connection lifecycle management, signer creation, auto-connect persistence, subscribable state for UI frameworks, and `signIn` (SIWS-as-connect). Tests are thorough and well-structured with a clean mock setup.

Overall this is solid work — the architecture is clean, the SSR guard is well thought out, and the test coverage is excellent. I have a few things to flag, most notably around the `signIn` error handling and some concurrency edge cases.

## Key Things to Watch

1. **`signIn` is missing error handling and status transition** — Unlike `connect()`, `signIn()` doesn't set `status: 'connecting'` before the async call, and has no try/catch to call `disconnectLocally()` on failure. If `signIn` throws, the store status is never reset.

2. **`setState` notifies listeners even when nothing changed** — The listener notification runs unconditionally outside the snapshot-change check. For `useSyncExternalStore` this is mostly fine (React does its own `Object.is` check), but other frameworks that re-render on every listener call will get unnecessary updates.

3. **No concurrency guard on `connect`/`signIn`** — Calling `connect(walletA)` then `connect(walletB)` before the first resolves can leave leaked event subscriptions from walletA, since `setConnected` for walletA runs and sets up `walletEventsCleanup` before walletB overwrites it.

4. **Late registration handler fires concurrent async IIFEs** — If multiple wallets register in quick succession, multiple `attemptSilentReconnect` calls can race.

## Notes for Subsequent Reviewers

- The persistence format `walletName:accountAddress` uses `lastIndexOf(':')` for parsing, which correctly handles wallet names containing colons (base58 addresses don't contain `:`). The malformed-key case is tested.
- The SSR stub returns a frozen singleton snapshot, which is important for `useSyncExternalStore` referential stability.
- The `walletAsPayer` plugin uses `Object.defineProperty` for a dynamic `payer` getter since `extendClient` freezes the result — clever approach.
- Auto-connect tests use fake timers to control the async flow — good pattern.

packages/kit-plugin-wallet/src/store.ts
mcintyre94: Fixed, now sets `connecting` and disconnects in a catch

packages/kit-plugin-wallet/src/store.ts
mcintyre94: Fixed, moved into the if

packages/kit-plugin-wallet/src/store.ts
mcintyre94: - Fixed - catch at the end sets `disconnected` state
- Concurrency is fixed with a `connectGeneration` shared between connect, signIn and reconnect. This lets us identify stale responses in all cases.

packages/kit-plugin-wallet/src/store.ts
mcintyre94: Fixed by adding `connectGeneration`, incremented in connect, signIn and auto connect.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,603 @@
...
+            if (!filterWallet(refreshed)) {
+                disconnectLocally();
+                return;
+            }
+
+            let updates: Partial<WalletStoreState> = { connectedWallet: refreshed };
+
```
trevor-cortex: **Bug — signer computed for stale account when both `accounts` and `features` change in the same event.**

`handleFeaturesChanged()` reads `state.account` (the _old_ account) because `setState` hasn't been called yet at this point. If `handleAccountsChanged` returned a new account + signer, the features handler overwrites the signer with one computed from the old account:

```
// handleAccountsChanged returns: { account: newAccount, signer: signerForNewAccount }
// handleFeaturesChanged reads state.account (old!) and returns: { signer: signerForOldAccount }
// Spread: signerForOldAccount wins → mismatch
```

Fix: either pass the resolved account through, or skip the features signer update when accounts also changed (since the account handler already creates a fresh signer):

```ts
if (properties.accounts) {
    const result = handleAccountsChanged(refreshed);
    if (result === null) {
        disconnectLocally();
        return;
    }
    updates = { ...updates, ...result };
}
if (properties.features && !properties.accounts) {
    updates = { ...updates, ...handleFeaturesChanged() };
}
```

Alternatively, have `handleFeaturesChanged` accept an account parameter.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,603 @@
...
+            const existingAccounts = [...uiWallet.accounts];
+
+            await connectFeature.connect();
+
+            // A newer connect/signIn was started while we were awaiting — bail.
+            if (generation !== connectGeneration) return [];
+
+            // Refresh UiWallet to get updated accounts after connect.
+            const refreshedWallet = refreshUiWallet(uiWallet);
+            const allAccounts = refreshedWallet.accounts;
```
trevor-cortex, line 419: Nit: `disconnect()` doesn't participate in the `connectGeneration` counter, so `disconnectLocally()` in the `finally` block can wipe out a `connect()` that completed while the `disconnect` feature call was in-flight.

Low priority — it requires the user to connect a new wallet while an old wallet's disconnect is still pending, which is unusual. But if you wanted to guard against it:

```ts
async function disconnect(): Promise<void> {
    if (!state.connectedWallet) return;
    const currentWallet = state.connectedWallet;
    const generation = connectGeneration; // snapshot
    setState({ status: 'disconnecting' });
    try {
        // ... disconnect feature call
    } finally {
        if (generation === connectGeneration) {
            disconnectLocally();
        }
    }
}
```

trevor-cortex (commented):
## Re-review

Great progress — all four issues from my first review have been addressed:

- ✅ `signIn` now has proper `status: 'connecting'` transition, try/catch, and `disconnectLocally()` on error
- ✅ `setState` only notifies listeners when snapshot-relevant fields actually change
- ✅ `connectGeneration` counter guards against concurrent connect/signIn races (with thorough tests)
- ✅ Late-registration reconnect handler has `.catch()` for error recovery

The concurrency test coverage is excellent — stale rejection, concurrent connects, cross-connect/signIn superseding, and auto-reconnect vs explicit connect are all well covered.

## Remaining items

I found one state consistency bug and a couple of smaller things:

1. **Signer/account mismatch when both `accounts` and `features` change in a single event** — `handleFeaturesChanged` reads `state.account` (the old account) instead of the potentially-updated account from `handleAccountsChanged`, so the signer can be computed for the wrong account. (See inline comment.)

2. **`disconnect()` finally block can clobber a concurrent `connect()`** — Since `disconnect()` doesn't participate in the generation counter, its `finally { disconnectLocally() }` can wipe out a `connect()` that completed while the disconnect was in flight. Low priority since it's an unusual user interaction pattern, but worth noting.

3. **No validation in `selectAccount`** — Any `UiWalletAccount` is accepted regardless of whether it belongs to the connected wallet. If a foreign account is passed, the persisted key will have a mismatched wallet name + address. Not a crash, but surprising behavior.

None of these are blockers — #1 is the most impactful but only triggers on simultaneous account+feature wallet events (uncommon). Overall the implementation is clean and well-tested.

trevor-cortex (commented):
## Re-review (3rd pass)

All 7 issues from the previous two reviews are now addressed with corresponding test coverage:

- ✅ `signIn` error handling and status transition (from review 1)
- ✅ `setState` only notifies on actual changes (from review 1)
- ✅ `connectGeneration` guards connect/signIn/disconnect concurrency (from review 1)
- ✅ Late-registration reconnect has `.catch()` recovery (from review 1)
- ✅ **Signer/account mismatch fixed** — `handleFeaturesChanged` is now skipped when accounts also changed in the same event, so the signer is always computed for the correct account (from review 2)
- ✅ **`disconnect()` participates in generation counter** — captures `connectGeneration` before the async call and checks it in `finally`, preventing a stale disconnect from clobbering a concurrent connect (from review 2)
- ✅ **`selectAccount` validates account belongs to wallet** — refreshes the wallet and checks the address exists in the account list before accepting (from review 2)

The new tests are thorough — `'disconnect does not clobber concurrent connect'`, `'uses new account signer when both accounts and features change'`, `'selectAccount throws for account not in wallet'`, and `'still reconnects if wallet appears after timeout'` all directly verify the fixes.

The wallet plugin architecture is clean: four plugin variants (`walletSigner`, `walletPayer`, `walletIdentity`, `walletWithoutSigner`) built on a shared `createPlugin` helper with `defineSignerGetter` for dynamic payer/identity getters. SSR stub, persistence, auto-reconnect with timeout, and the concurrency model are all solid.

LGTM 🚢

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,622 @@
...
+
 // -- Store ------------------------------------------------------------------
 
 /** @internal */
-export function createWalletStore(_config: WalletPluginConfig): WalletStore {
-    throw new Error('not implemented');
+export function createWalletStore(config: WalletPluginConfig): WalletStore {
+    // -- SSR guard: on the server, return an inert stub ---------------------
+
+    if (!__BROWSER__) {
```
lorisleiva: Is that the desired behaviour for `__REACTNATIVE__` as well?

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,622 @@
...
+        } catch {
+            // Wallet doesn't support signing (read-only / watch wallet, or a
+            // non-Solana chain).
+            return null;
+        }
+    }
+
+    // -- Wallet discovery --------------------------------------------------
+
+    const registry = getWallets();
```
lorisleiva, line 199: Does that mean wallets are only discovered once when the plugin is installed? Should we offer a way to refresh the wallet list?

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,622 @@
...
+                .filter(filterWallet),
+        );
+    }
+
+    setState({ wallets: buildWalletList() });
+
+    // Listen for new wallets being registered
+    const unsubRegister = registry.on('register', () => {
+        setState({ wallets: buildWalletList() });
+    });
```
lorisleiva: Oh I see we have listeners sorry. Is it normal that `getWallets` is outside of `buildWalletList` though? It seems as though the wallets would be refreshed with the current implementation but I must be missing something.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,622 @@
...
+                      })
+                    : null,
+            status: s.status,
+            wallets: s.wallets,
+        });
+    }
+
+    let snapshot: WalletState = deriveSnapshot(state);
+
+    function setState(updates: Partial<WalletStoreState>): void {
```
lorisleiva: nit: maybe this internal helper should be called `updateState` since it doesn't fully set the whole object?

lorisleiva (approved):
Damn that's a lot of code people won't need to write for managing wallets haha. Thanks for that! I've got a few noob questions but ultimately you have more experience in the wallet space so I trust your judgement!

packages/kit-plugin-wallet/src/store.ts
mcintyre94: This is a bit confusing, but `getWallets()` is returning an interface called `Wallets`, which is basically a registry of wallets, rather than the wallets themselves:

https://github.com/wallet-standard/wallet-standard/blob/8418e7c5a152a4aa8c8c7f41c3e6c2c701ddf910/packages/core/app/src/wallets.ts#L69-L106

That interface has `.get()` which is what we call in `buildWalletList`, and it also has `.on()` to listen for changes.

So to relate it to the rest of our architecture, `getWallets()` is giving us a reactive store, and then `buildWalletList` and other calls to `registry.get()` are giving us a current view of the registered wallets. 

`getWallets()` is also memoized:

https://github.com/wallet-standard/wallet-standard/blob/8418e7c5a152a4aa8c8c7f41c3e6c2c701ddf910/packages/core/app/src/wallets.ts#L41-L43

lorisleiva: Ahhh okay, thanks for the explanation! Probably should have been called something like `getWalletRegistry` I guess but not a big deal.

mcintyre94: Agreed - added a comment for next time I forget how this works!

.changeset/busy-clowns-stick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solana/kit-plugin-wallet': minor
+---
+
+Add `walletSigner`, `walletIdentity`, `walletPayer`, and `walletWithoutSigner` plugins for framework-agnostic wallet management using wallet-standard. Provides wallet discovery, connection lifecycle, signer creation, auto-connect persistence, subscribable state for UI frameworks, and dynamic payer/identity integration.
```
mcintyre94: Note to self: make this changeset more of a new plugin announcement, with code examples

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -1,14 +1,828 @@
...
+        }
+        // Reject if this wallet is disconnecting
+        if (state.connectedWallet.name === disconnectingWalletName) {
+            throw new SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED, { operation: 'selectAccount' });
+        }
+        const refreshed = refreshUiWallet(state.connectedWallet);
+        const selectedAccount = refreshed.accounts.find(a => a.address === account.address);
+        if (!selectedAccount) {
+            // Note: no Kit SolanaError for this case
+            throw new Error(`Account ${account.address} is not available in wallet "${state.connectedWallet.name}"`);
```
mcintyre94: I think this case is worth adding to the Kit wallet errors before shipping

mcintyre94 (commented):
Dropping to draft for now - one last tidy up before shipping

PR #192, opened 2026-04-14
Remove `payer` config option from transaction planner plugins

lorisleiva:
This PR removes the `Partial<ClientWithPayer>` pattern from the transaction planner plugins. The payer must now be provided on the client (e.g. via a payer plugin) rather than passed as a config override, aligning with the composable plugin model.

PR #193, opened 2026-04-14
Replace `priorityFees` with `TransactionPlannerConfig` in transaction planner plugins

lorisleiva:
This PR replaces the inline `priorityFees` config on `rpcTransactionPlanner` and `litesvmTransactionPlanner` with a new `TransactionPlannerConfig` type that supports `microLamportsPerComputeUnit` and `version`. The all-in-one plugins (`solanaRpc` and `litesvm`) gain a `transactionConfig` option for passing this configuration through.

This prepares us for v1 transactions, which will configure priority fees in a completely different way (flat lamport amount instead of micro-lamports per compute unit). The `TransactionPlannerConfig` type gives us a place to provide per-version configuration — e.g. a future `{ version: 1, priorityFees: lamports(100n) }` variant — without breaking the existing API surface.

### DIFF
--- a/packages/kit-plugin-litesvm/src/index.browser.ts
+++ b/packages/kit-plugin-litesvm/src/index.browser.ts
@@ -1,4 +1,5 @@
 export type { FailedTransactionMetadata, TransactionMetadata } from 'litesvm';
+export type { LiteSvmConfig } from './litesvm';
 export type { LiteSVM } from './litesvm-connection';
 export type { LiteSvmRpcApi } from './litesvm-to-rpc';
 
--- a/packages/kit-plugin-litesvm/src/transaction-planner.ts
+++ b/packages/kit-plugin-litesvm/src/transaction-planner.ts
@@ -32,22 +32,17 @@ import {
  *     .use(litesvmTransactionPlanExecutor());
  * ```
  */
-export function litesvmTransactionPlanner(
-    config: {
-        /**
-         * The priority fees to be set on the transaction in micro lamports per compute unit.
-         * Defaults to using no priority fees.
-         */
-        priorityFees?: MicroLamports;
-    } = {},
-) {
+export function litesvmTransactionPlanner(config: TransactionPlannerConfig = {}) {
     return <T extends ClientWithPayer>(client: T) => {
         const transactionPlanner = createTransactionPlanner({
             createTransactionMessage: () => {
                 return pipe(
-                    createTransactionMessage({ version: 0 }),
+                    createTransactionMessage({ version: config.version ?? 0 }),
                     tx => setTransactionMessageFeePayerSigner(client.payer, tx),
-                    tx => (config.priorityFees ? setTransactionMessageComputeUnitPrice(config.priorityFees, tx) : tx),
+                    tx =>
+                        config.microLamportsPerComputeUnit
+                            ? setTransactionMessageComputeUnitPrice(config.microLamportsPerComputeUnit, tx)
+                            : tx,
                 );
             },
         });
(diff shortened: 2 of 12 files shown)


lorisleiva:
* **#193** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/193?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#192** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/kit-plugin-litesvm/src/transaction-planner.ts
```diff
@@ -32,26 +32,44 @@ import {
...
+     * The priority fees to be set on the transaction in micro lamports per compute unit.
+     * Defaults to using no priority fees.
+     */
+    microLamportsPerComputeUnit?: MicroLamports;
+    /**
+     * The transaction message version to use when creating transaction messages.
+     * Defaults to version 0.
+     */
+    version?: 'legacy' | 0;
+};
```
trevor-cortex, line 75: Nit: since this type is identical to the one in `kit-plugin-rpc/src/transaction-planner.ts`, would it make sense to extract it into a shared package (e.g. `kit-plugin-instruction-plan` or a new shared types package) to avoid the manual sync + the explicit disambiguation in `kit-plugins/src/index.ts`? Not blocking — the TODO comments and barrel re-export handle it fine for now.

packages/kit-plugin-litesvm/src/index.browser.ts
```diff
@@ -1,4 +1,5 @@
 export type { FailedTransactionMetadata, TransactionMetadata } from 'litesvm';
+export type { LiteSvmConfig } from './litesvm';
```
trevor-cortex, line 2: Good catch exporting `LiteSvmConfig` from the browser entry point so consumers can still reference the type even when the runtime function throws.

trevor-cortex (commented):
## Summary

Clean refactor that replaces the inline `priorityFees` option on `rpcTransactionPlanner` and `litesvmTransactionPlanner` with a new `TransactionPlannerConfig` type supporting `microLamportsPerComputeUnit` and `version`. The all-in-one plugins (`solanaRpc` / `litesvm`) gain a `transactionConfig` passthrough option. This sets up a nice extension point for future v1 transaction support with different priority fee semantics.

## Notes

- The `TransactionPlannerConfig` type is defined identically in both `kit-plugin-rpc` and `kit-plugin-litesvm` transaction-planner modules. The barrel export in `kit-plugins` correctly disambiguates by explicitly re-exporting from `kit-plugin-rpc`. This is fine for now, but worth noting that the two definitions need to stay in sync manually (the TODO comments help here).
- `kit-client-rpc` still exposes a `priorityFees` field on its `ClientConfig` type and maps it to `microLamportsPerComputeUnit` internally — this is the right call since `createClient`/`createLocalClient` are already deprecated.
- The RPC transaction planner still includes `fillTransactionMessageProvisoryComputeUnitLimit` in its pipeline while the LiteSVM planner does not — this is pre-existing and intentional (LiteSVM doesn't need CU estimation), just noting for subsequent reviewers.
- Test coverage is thorough: version default, legacy version, compute unit price presence/absence.

Looks good to me 👍

packages/kit-plugin-litesvm/src/transaction-planner.ts
lorisleiva: I did that on purpose to allow the two config objects to grow independently of each other.

mcintyre94 (approved):
This looks great! It'll be really nice to get the simpler priority fees in v1 here too.

PR #199, opened 2026-04-20
Add readonly variants of SolanaRpc plugins

mcintyre94:
This PR adds a `solanaRpcReadOnly` plugin, as well as cluster-specific variants. This includes the RPC, RPC subscriptions, getMinimumBalance, and airdrop (on appropriate clusters). It does not include transaction planning/sending, and therefore does not require a payer.

It's defined as a strict subset of the `solanaRpc` plugin, which is refactored to compose from the readonly variant to avoid drift. Similarly its config is a subset of the `solanaRpc` config, excluding fields related to transaction sending.

This is useful for a range of apps/scripts that use Solana data but do not need to themselves send transactions. I came across the idea while speccing a react library, but figured it makes sense as an independent plugin too.

### DIFF
--- a/.changeset/ten-kings-hug.md
+++ b/.changeset/ten-kings-hug.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-rpc': minor
+---
+
+Add a read-only RPC pathway: `solanaRpcReadOnly()` plus the cluster convenience wrappers `solanaMainnetRpcReadOnly()`, `solanaDevnetRpcReadOnly()`, and `solanaLocalRpcReadOnly()` (and the matching `SolanaRpcReadOnlyConfig` type). These install `client.rpc`, `client.rpcSubscriptions`, and `client.getMinimumBalance` without the transaction-planning or transaction-sending plugins, so they have no `payer` prerequisite. The devnet and local variants also install `rpcAirdrop` (airdrop does not require a payer). Use for dashboards, explorers, on-chain watchers, and other read-only flows.
--- a/packages/kit-plugin-rpc/README.md
+++ b/packages/kit-plugin-rpc/README.md
@@ -120,6 +120,97 @@ _See `solanaRpc` for available features, plus:_
     await client.airdrop(address('HQVxiMVDoV9jzG4tpoxmDZsNfWvaHXm8DGGv93Gka75v'), lamports(1_000_000_000n));
     ```
 
+## `solanaRpcReadOnly` plugin
+
+The `solanaRpcReadOnly` plugin sets up a read-only Solana RPC client in a single call. It installs an RPC connection, RPC Subscriptions, and minimum balance computation on the client, but does not install transaction planning or transaction sending.
+
+Unlike `solanaRpc`, this plugin does not require a `payer` on the client, making it ideal for dashboards, explorers, on-chain watchers, and other read-only flows.
+
+### Installation
+
+```ts
+import { createClient } from '@solana/kit';
+import { solanaRpcReadOnly } from '@solana/kit-plugin-rpc';
+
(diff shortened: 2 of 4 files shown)


mcintyre94:
@trevor-cortex

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds a read-only pathway to `@solana/kit-plugin-rpc`: `solanaRpcReadOnly()` plus the three cluster-specific wrappers (`solanaMainnetRpcReadOnly`, `solanaDevnetRpcReadOnly`, `solanaLocalRpcReadOnly`) and a matching `SolanaRpcReadOnlyConfig` type. The existing `solanaRpc` is nicely refactored to compose from `solanaRpcReadOnly` so the read/write variant stays a strict superset — no drift between configs, no drift between plugin pipelines. Devnet and local read-only variants install `rpcAirdrop` (no payer needed); mainnet read-only intentionally does not. Tests cover the new public surface across all four variants including negative assertions that transaction planning/sending are absent.

Overall the shape of the change is clean and the composition is exactly what I'd hope for. One thing to address before this lands, plus a few nits.

## Should address

- **README is missing the new APIs.** The repo's `CLAUDE.md` is explicit: *"When adding a new public API, add a section to the package's `README.md` following the existing structure."* This PR adds four new public functions and one new exported type, none of which appear in `packages/kit-plugin-rpc/README.md`. Please add sections for `solanaRpcReadOnly`, `solanaMainnetRpcReadOnly`, `solanaDevnetRpcReadOnly`, and `solanaLocalRpcReadOnly` mirroring the existing `solanaRpc` / `solanaMainnetRpc` / etc. sections, and call out the key distinction (no `payer` prerequisite, no transaction planning/sending).

## Nits

- The subscription-URL tests for `solanaRpcReadOnly` (`derives the WebSocket URL from the RPC URL by default` and `accepts an explicit rpcSubscriptionsUrl`) only assert that `client.rpcSubscriptions` exists, which is the same assertion as the first test — they don't actually verify URL wiring. This matches the shape of the existing `solanaRpc` tests, so it's consistent with repo style, but if you wanted to tighten it you could spy on `createSolanaRpcSubscriptions` or otherwise verify the URL. Not a blocker.
- The JSDoc on `solanaDevnetRpcReadOnly` says *"so local dev / test flows can request SOL from the faucet"* — the "local dev" phrasing reads slightly awkwardly on the *devnet* variant (as opposed to the local variant just below it). Minor wording tweak.
- The changeset is a `minor` bump, which is correct for a purely additive feature. Noting only that this package is pre-1.0 and the repo's changeset guidance says consumers may treat minors as potentially breaking — there's no real breakage here though.

- The `<T extends object>` constraint on all four readonly wrappers (vs. `<T extends ClientWithPayer>` on the write variants) is the core correctness claim of the PR and it holds up: none of the plugins composed inside `solanaRpcReadOnly` (`solanaRpcConnection`, `solanaRpcSubscriptionsConnection`, `rpcGetMinimumBalance`, `rpcAirdrop`) require a payer.
- Worth sanity-checking that consumers importing `SolanaRpcConfig` still get the same shape after the refactor — the intersection `SolanaRpcReadOnlyConfig<TClusterUrl> & { maxConcurrency?, skipPreflight?, transactionConfig? }` should be structurally identical to the prior definition, and I believe it is, but type-level test (e.g. via `expect-type`) would catch any accidental divergence here and on the new `SolanaRpcReadOnlyConfig` type.

lorisleiva:
@trevor-cortex Just testing something (unrelated to this PR). Can you tell me in one or two lines if this is okay to merge?

trevor-cortex:
Yep, looks good to merge — README is now covered for all four new functions and the `SolanaRpcReadOnlyConfig` type, and the remaining items from my earlier review were all nits rather than blockers.

mcintyre94:
Closing in favour of #201

PR #201, opened 2026-04-21
Merge `rpc` and `rpcSubscriptions` setup into `solanaRpcConnection`

lorisleiva:
This PR merges the `rpc` and `rpcSubscriptions` setup into a single `solanaRpcConnection` plugin. The purpose of this change is to have a convenient payer-free version of the `solanaRpc` plugin that isn't set up to send transactions.

The plugin now takes a `SolanaRpcConnectionConfig` object (instead of positional `(url, rpcConfig?)` arguments) and installs both `client.rpc` and `client.rpcSubscriptions` in one call, deriving the WebSocket URL from the RPC URL when `rpcSubscriptionsUrl` is not provided. `SolanaRpcConfig` is reshaped to extend `SolanaRpcConnectionConfig` so the URL and transport options live in one place.

As a consequence, three helpers become redundant and are soft-deprecated with proper migration guidance in their docblocks: `solanaRpcSubscriptionsConnection` (subscriptions are now installed by `solanaRpcConnection` itself), and the generic `rpcConnection` / `rpcSubscriptionsConnection` wrappers (trivial one-liners around `extendClient`).

Docblocks, tests, and READMEs across the affected packages are updated to use the new single-call form. Tests now also cover the derived WebSocket URL, an explicit `rpcSubscriptionsUrl`, and forwarded `rpcConfig` / `rpcSubscriptionsConfig`.


lorisleiva:
* **#201** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

.changeset/heavy-papers-cut.md
```diff
@@ -0,0 +1,22 @@
+---
+'@solana/kit-plugin-rpc': minor
```
trevor-cortex, line 2: Consider calling out the breaking nature of this change explicitly. `solanaRpcConnection('url')` no longer compiles — existing call sites will fail to build after upgrading. A leading "**Breaking:**" note, or bumping to `major`, would make this easier to catch in changelogs. (Pre-1.0 `minor` is acceptable per `CLAUDE.md`, but the body currently reads as an additive refactor.)

packages/kit-plugin-rpc/src/solana-rpc.ts
```diff
@@ -211,55 +231,84 @@ export function solanaLocalRpc(config?: Partial<SolanaRpcConfig<string>>) {
...
+ * import { createClient, mainnet } from '@solana/kit';
+ * import { solanaRpcConnection } from '@solana/kit-plugin-rpc';
+ *
+ * const client = createClient().use(
+ *     solanaRpcConnection({ rpcUrl: mainnet('https://api.mainnet-beta.solana.com') }),
+ * );
+ * ```
+ *
+ * @see {@link SolanaRpcConnectionConfig}
+ * @see {@link solanaRpc}
```
trevor-cortex, line 268: The explicit return type annotation here is inconsistent with every other connection helper in this file (`rpcConnection`, `rpcSubscriptionsConnection`, the legacy `rpc`), which all rely on `extendClient`'s inferred return type. If the annotation is load-bearing (e.g. `extendClient` loses `SolanaRpcApiFromClusterUrl<TClusterUrl>` narrowing through inference), a brief comment explaining why would help future maintainers. If it isn't load-bearing, dropping it keeps the file stylistically consistent.

packages/kit-plugin-rpc/test/index.test.ts
```diff
@@ -31,16 +31,48 @@ describe('rpcSubscriptionsConnection', () => {
...
-        const client = createClient().use(solanaRpcConnection(mainnet('https://api.mainnet-beta.solana.com')));
+        const client = createClient().use(
+            solanaRpcConnection({ rpcUrl: mainnet('https://api.mainnet-beta.solana.com') }),
+        );
         expectTypeOf(client.rpc).not.toHaveProperty('requestAirdrop');
     });
+
+    it('derives the WebSocket URL from the RPC URL by default', () => {
+        const client = createClient().use(solanaRpcConnection({ rpcUrl: 'https://api.mainnet-beta.solana.com' }));
+        expect(client).toHaveProperty('rpcSubscriptions');
```
trevor-cortex: These two cases (derived WS URL and explicit `rpcSubscriptionsUrl`) only check that `rpcSubscriptions` was installed, which the earlier test already covers. The interesting behavior — that the correct URL was passed to `createSolanaRpcSubscriptions` — isn't actually asserted. Consider spying on `createSolanaRpcSubscriptions` (or reaching into the returned instance) to assert the resolved URL, so the derivation logic is locked down by the test.

trevor-cortex (commented):
## Summary

Merges `solanaRpcConnection` and `solanaRpcSubscriptionsConnection` into a single plugin that takes a `SolanaRpcConnectionConfig` object and installs both `client.rpc` and `client.rpcSubscriptions` in one call, deriving the WS URL from the RPC URL when not provided. `SolanaRpcConfig` is refactored to extend `SolanaRpcConnectionConfig` so URL/transport options live in one place. Three helpers are soft-deprecated with migration guidance: `solanaRpcSubscriptionsConnection`, `rpcConnection`, and `rpcSubscriptionsConnection`. Docblocks, READMEs, and tests across the affected packages are updated consistently.

The ergonomics win is real — most consumers only ever paired these two plugins and dealt with the ws/wss URL boilerplate. Nice cleanup.

## Things to watch out for

- **This is a breaking signature change to `solanaRpcConnection` with no runtime deprecation alias.** The old `solanaRpcConnection('https://...')` call becomes a compile-time error. The changeset is `minor`, which is fine pre-1.0 per `CLAUDE.md` ("`minor` bumps may be treated as breaking"), but the changeset body doesn't currently flag it as breaking — users scanning changelogs for breaking changes may miss it. Consider either bumping to `major`, or making the breaking nature more explicit in the changeset intro (e.g. "**Breaking:** `solanaRpcConnection` now takes a config object instead of a positional URL").
- The new `solanaRpcConnection` adds an explicit return type (`Omit<T, 'rpc' | 'rpcSubscriptions'> & { rpc: ...; rpcSubscriptions: ... }`) while the sibling helpers (`rpcConnection`, `rpcSubscriptionsConnection`, the legacy `rpc`) all rely on inferred return types from `extendClient`. Worth verifying this is intentional and not a hint that `extendClient` inference is lossy for one of these callers.
- The new tests cover the *presence* of `rpcSubscriptions` for the "derived URL" and "explicit `rpcSubscriptionsUrl`" cases but don't actually assert that the URL is wired correctly. Not a blocker since both paths converge on `createSolanaRpcSubscriptions`, but a spy/mock on `createSolanaRpcSubscriptions` asserting the resolved URL would make the core new behavior a tested guarantee rather than a code-reading guarantee.

## Notes for reviewers

- README/docblock updates across all packages are thorough and consistent — verified the before/after examples all compile with the new signature.
- The `config.rpcUrl.replace(/^http/, 'ws')` derivation correctly handles both `http→ws` and `https→wss` (the latter because `https` starts with `http` → `wss`). No regression from the prior implementation in `solanaRpc`, which used the same regex.
- `@solana/kit-plugins` umbrella README is doc-only and correctly has no changeset.

packages/kit-plugin-rpc/src/solana-rpc.ts
lorisleiva: When not providing an explicit return type here, TypeScript was getting confused on subsequent plugins for consumers of this plugin. I believe this is mainly because the Solana RPC factories use a lot of complex types to infer the Cluster, API and transports types on top of all the nested `Omit<T, ...>` that `extendClient` introduces.

Being more explicit about the return type here (without needing an explicit cast btw) helpers TypeScript by not needing to infer so much when chaining RPC plugins.

.changeset/heavy-papers-cut.md
lorisleiva: 0.x so `minor` is okay for breaking changes. The changeset is detailed enough IMO.

packages/kit-plugin-rpc/test/index.test.ts
lorisleiva: Good shout, I updated the tests to include some mocks.

.changeset/heavy-papers-cut.md
```diff
@@ -0,0 +1,22 @@
...
+---
+
+Merge the `rpc` and `rpcSubscriptions` setup into a single `solanaRpcConnection` plugin.
```
trevor-cortex, line 5: Consider prefixing this line with `**Breaking:**` (or bumping the package to `major`). The positional-arg → config-object change is a compile-time break for any existing caller of `solanaRpcConnection('url')`, and readers scanning the changelog for breaking changes won't infer that from a `minor` bump with a neutral headline. Per this repo's `CLAUDE.md`, pre-1.0 minor bumps may be treated as breaking, but an explicit callout here would help downstream consumers.

trevor-cortex (commented):
## Re-review

Thanks for the test improvements — the URL derivation concern from my last pass is fully addressed. The new mock-based tests on `createSolanaRpc` / `createSolanaRpcSubscriptions` (covering `https→wss`, `http→ws`, explicit `rpcSubscriptionsUrl`, and forwarded `rpcConfig` / `rpcSubscriptionsConfig`) lock down the core new behavior cleanly. 👍

## Still open from the previous review

- **Changeset doesn't flag the breaking signature change** — see inline comment on the changeset. `solanaRpcConnection('url')` → `solanaRpcConnection({ rpcUrl: 'url' })` is a compile-time break and it's not currently called out in the changeset body.
- **Explicit return type on `solanaRpcConnection` vs. inferred on siblings.** The new `solanaRpcConnection` in `packages/kit-plugin-rpc/src/solana-rpc.ts` spells out its return type as `Omit<T, 'rpc' | 'rpcSubscriptions'> & { rpc: Rpc<SolanaRpcApiFromClusterUrl<TClusterUrl>>; rpcSubscriptions: RpcSubscriptions<SolanaRpcSubscriptionsApi> }`, while `rpcConnection`, `rpcSubscriptionsConnection`, and the legacy `rpc` all rely on inferred return types from `extendClient`. Worth confirming this annotation is load-bearing (e.g. because `extendClient` inference drops the narrowed `SolanaRpcApiFromClusterUrl<TClusterUrl>` for this specific caller) vs. removable. If it's load-bearing, a one-line comment explaining why would help future readers; if not, dropping it keeps the style consistent with the sibling helpers.

## New minor observation

- In the `solanaRpc` describe block in `packages/kit-plugin-rpc/test/index.test.ts`, the new `'derives the WebSocket URL from the RPC URL by default'` and `'accepts an explicit rpcSubscriptionsUrl'` tests only assert `client` has `rpcSubscriptions`, which is already covered by the `'sets up a full RPC client with all plugins'` test above them. Since `solanaRpc` now delegates to `solanaRpcConnection` (whose URL derivation is thoroughly tested a few `describe`s up), these are either redundant or should be strengthened with assertions on the mocked `createSolanaRpcSubscriptions` call — e.g. `expect(createSolanaRpcSubscriptions).toHaveBeenCalledWith('wss://api.mainnet-beta.solana.com', ...)` for the default case and `'wss://custom-ws.solana.com'` for the explicit one.

None of these are blockers — leaving this as `COMMENT` again.

mcintyre94 (approved):
Nice, this looks great! Much cleaner + solves for a single plugin to get read-only RPC + subscriptions. I'll close the `RpcReadOnly` PR in favour of this because this covers what we need.

PR #204, opened 2026-04-24
test: add signer override tests

amilz:
Adds regression tests covering `extendClient` override behavior - chaining two plugins that set the same client field (e.g. `signer()` then `payer()`).

All three tests currently throw `TypeError: Cannot redefine property: <field>` and will pass once the upstream fix lands.

### Depends on

- anza-xyz/kit#1569 - fixes `extendClient` to allow later plugins to override keys set by earlier plugins

### Test plan

- [X] Tests fail on current `@solana/plugin-core@6.8.0` (reproduces the bug)
- [X] Tests pass against anza-xyz/kit#1569 (verified locally via `pnpm patch`)

### DIFF
--- a/packages/kit-plugin-signer/test/signer.test.ts
+++ b/packages/kit-plugin-signer/test/signer.test.ts
@@ -31,3 +31,35 @@ describe('identity', () => {
         expect(client.identity).toBe(mySigner);
     });
 });
+
+describe('override behavior', () => {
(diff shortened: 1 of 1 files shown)


amilz:
@lorisleiva rebased this on top of the 6.9 commit, so i think CI should pass now.

lorisleiva (approved):
Thanks! Sorry your last message got buried in my GH notifications.

PR #210, opened 2026-05-05
fix(kit-plugin-rpc): derive ws://...:8900 for local validator rpcUrl

amilz:
## Summary

`solanaRpcConnection({ rpcUrl })` derived `rpcSubscriptionsUrl` by swapping protocol only, so `http://127.0.0.1:8899` produced `ws://127.0.0.1:8899` — the wrong port for `solana-test-validator`, whose WebSocket listens on `8900`. Subscriptions silently failed to connect.

This patch routes derivation through an exact-string lookup table (`LOCAL_VALIDATOR_RPC_SUBSCRIPTIONS_URLS`) that maps the two canonical local validator URLs (`http://127.0.0.1:8899`, `http://localhost:8899`) to their matching `:8900` subs URLs. Every other URL falls through to the existing protocol swap, so user-customized validator setups are untouched. `solanaLocalRpc` and the deprecated `localhostRpc` reference the same constants so defaults can't drift from the lookup table.

Introduced in #201. Tracked in [DEV-489](https://linear.app/solana-fndn/issue/DEV-489).

cc @lorisleiva for review.

## Test plan

- [x] `pnpm test` in `packages/kit-plugin-rpc` (144 passed)
- [x] `pnpm lint`
- [x] `pnpm build`
- [x] Verified live against running `solana-test-validator`: both `http://127.0.0.1:8899` and `http://localhost:8899` derive working subs URLs and receive `slotNotifications`
- [x] New unit tests pin: 8899→8900 rewrite for both `127.0.0.1` and `localhost`; `https://127.0.0.1:8899` left untouched (exact-string http only); non-default ports preserved; `mainnet(...)` derivation unchanged; deprecated `rpc()` factory exercises the same helper

### DIFF
--- a/.changeset/huge-lands-argue.md
+++ b/.changeset/huge-lands-argue.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-rpc': patch
+---
+
+Fix `solanaRpcConnection` WebSocket URL derivation for the canonical local validator endpoint. When `rpcUrl` is `http://127.0.0.1:8899` or `http://localhost:8899`, the derived `rpcSubscriptionsUrl` now correctly targets port `8900` to match `solana-test-validator` defaults. Non-default ports and remote endpoints are unchanged.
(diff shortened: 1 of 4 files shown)


lorisleiva (approved):
Thanks! That's a lot more robust. 👌

PR #281, opened 2026-06-19
Refactor plan executor to use `estimateResourceLimits` functions

mcintyre94:
This PR refactors `kit-plugin-rpc` to use the `estimateResourceLimits` functions from Kit instead of the `estimateComputeUnitLimits` ones.

For legacy and v0 transactions the behavior is unchanged, these set compute unit limit exactly the same as the previous helpers. But for v1 transactions we now additionally estimate and set the `loadedAccountsDataSizeLimit` resource.

This will allow users of the default transaction planner/executor to send v1 transactions when those are enabled.

Note that `TransactionPlannerConfig.version` still not does include v1, as the Kit `createTransactionMessage` type doesn't. This is just preparatory for that version being available.

### DIFF
--- a/.changeset/eager-coats-tie.md
+++ b/.changeset/eager-coats-tie.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-rpc': minor
+---
+
+Use Kit's resource limit estimation helpers in the RPC transaction planner and executor. In addition to the compute unit limit, the executor now estimates the loaded accounts data size limit for version 1 transactions, and the planner fills provisory resource limits accordingly. Behavior is unchanged for legacy and version 0 transactions.
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ yarn-error.log*
 test-ledger/
 target/
 CHANGELOG.md
+
+# LLM
+.claude/worktrees/
--- a/packages/kit-plugin-rpc/README.md
+++ b/packages/kit-plugin-rpc/README.md
@@ -222,7 +222,7 @@ const client = createClient()
 
 ## `rpcTransactionPlanner` plugin
 
-This plugin provides a default transaction planner that creates transaction messages with a fee payer, a provisory compute unit limit, and optional priority fees.
+This plugin provides a default transaction planner that creates transaction messages with a fee payer, provisory resource limits (a compute unit limit, plus a loaded accounts data size limit for version 1 transactions), and optional priority fees.
 
 ### Installation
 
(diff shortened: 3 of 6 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Migrates `kit-plugin-rpc`'s default planner/executor from Kit's `estimateComputeUnitLimit*` helpers to the newer `estimateResourceLimits*` family. For legacy and v0 transactions behavior is unchanged. For v1 transactions, the executor now also estimates and sets `loadedAccountsDataSizeLimit`, and the planner fills provisory resource limits via `fillTransactionMessageProvisoryResourceLimits` (the v1 variant of the old CU-only helper). As the PR notes, `TransactionPlannerConfig.version` still doesn't expose v1 — this is preparatory plumbing for when Kit's `createTransactionMessage` types catch up.

The refactor is nicely clean: the local `needsComputeUnitEstimation` + `MAX_COMPUTE_UNIT_LIMIT` machinery is gone, replaced by delegating gating to Kit's `estimateAndSetResourceLimitsFactory` (which calls our estimator only when something actually needs estimating). The `didSimulateToEstimate` flag is captured via a closure-injected `onSimulate` callback, which preserves the original skip-preflight-on-estimation behavior. The `skipPreflight` recovery path correctly forwards `error.context.loadedAccountsDataSize` when present.

## Things to watch / verify

- **Recovery path for v1 is untested.** The new conditional in the `catch` block — `error.context.loadedAccountsDataSize == null ? { computeUnitLimit } : { computeUnitLimit, loadedAccountsDataSizeLimit: ... }` — has no test coverage. The existing `skipPreflight + failed simulation` tests only exercise v0 (no `loadedAccountsDataSize`). A v1 variant of that test would cover the new branch. Non-blocking, but worth adding.
- **Buffer applies to CU only.** The 10% buffer is intentionally not applied to `loadedAccountsDataSizeLimit` (data size from simulation is exact, unlike CUs which vary). Worth a sentence in the docstring of `bufferAndRecoverResourceLimits` to make that intent explicit, since the function name suggests it buffers everything.
- **`onSimulate()` fires before the try.** That means `didSimulateToEstimate` becomes `true` even when the estimator throws non-recoverable errors. That's correct in the current flow (a non-recoverable throw propagates and the send never happens), but it's a subtle coupling — anything that ever catches and continues past `estimateAndSetResourceLimits` in the future would see a stale `true`. Not a problem today, just noting.
- **`.gitignore` is missing a trailing newline** (`.claude/worktrees/\ No newline at end of file`). Trivial nit, but worth fixing while you're here.
- **Changeset bump.** `minor` looks right: new v1-only behavior, no breaking changes to existing v0/legacy paths.

- The `as ResourceLimitsEstimate<typeof transactionMessage>` casts are needed because the conditional return-shape can't be inferred statically from the input message version. Reasonable concession.
- README table updated consistently with the new "every applicable resource limit" wording — matches the docstring on `skipPreflight`. Good.
- The three new tests use `@ts-expect-error` on `createTransactionMessage({ version: 1 })`, consistent with the PR description (v1 not yet in Kit's public types).

LGTM overall — the only thing I'd actually act on is the v1 recovery-path test; everything else is style/nitpick.

mcintyre94:
> Recovery path for v1 is untested. 

Test added

> Buffer applies to CU only.

Intentional, doc added

`onSimulate() fires before the try.`

Agreed, moved this after the simulate

lorisleiva (approved):
Nice, thank you!

PR #290, opened 2026-06-29
Add RPC resource limit estimation opt-out

catmcgee:
Adds `resourceLimitEstimation: "estimate" | "none"` to the RPC planner/executor path.

By default, behavior is unchanged: the planner reserves provisional resource-limit instructions and the executor estimates/sets missing limits before sending

When set to `none`:
- `rpcTransactionPlanner` does not add provisional resource-limit instructions
- `rpcTransactionPlanExecutor` does not simulate to estimate or inject missing limits
- explicit resource limits already present in a transaction message are preserved
- `solanaRpc` passes the same setting to both planner and executor

This is useful for transactions that are close to the message size limit, where adding a compute-budget instruction can make an otherwise valid transaction too large

### DIFF
--- a/packages/kit-plugin-rpc/README.md
+++ b/packages/kit-plugin-rpc/README.md
@@ -44,6 +44,7 @@ All options are provided via a `SolanaRpcConfig` object:
 - `transactionConfig`: Options to configure how transaction messages are created. See `rpcTransactionPlanner` options below.
 - `maxConcurrency`: Maximum number of concurrent transaction executions. Defaults to 10.
 - `skipPreflight`: Whether to always skip preflight simulation. Defaults to `false`.
+- `resourceLimitEstimation`: Whether to reserve, estimate, and set resource limits. Accepts `'estimate'` or `'none'`. Defaults to `'estimate'`.
 
 ### Features
 
@@ -246,6 +247,7 @@ All options are provided via a `TransactionPlannerConfig` object:
 
 - `version`: The transaction message version to use. Accepts `0` or `'legacy'`. Defaults to `0`.
 - `microLamportsPerComputeUnit`: Priority fees in micro lamports per compute unit. Defaults to no priority fees.
+- `resourceLimitEstimation`: Whether to reserve space for resource limits that the executor can later estimate and set. Accepts `'estimate'` or `'none'`. Defaults to `'estimate'`.
 
 ### Features
 
@@ -278,6 +280,7 @@ const client = await createClient()
 
 - `maxConcurrency`: Maximum number of concurrent executions (default: 10).
 - `skipPreflight`: Whether to skip the preflight simulation when sending transactions (default: `false`).
+- `resourceLimitEstimation`: Whether to estimate and set missing resource limits before sending. Accepts `'estimate'` or `'none'`. Defaults to `'estimate'`.
 
 ### Features
 
@@ -300,6 +303,8 @@ Setting `skipPreflight: true` changes the behavior:
 | Estimation succeeds | Set limits, skip preflight       | Set limits, skip preflight             |
 | Estimation fails    | Throw                            | Use consumed resources, skip preflight |
 | Explicit limits set | Run preflight                    | Skip preflight                         |
+
+Set `resourceLimitEstimation: 'none'` to opt out of automatic resource-limit handling. The RPC planner will not reserve provisional resource-limit instructions, and the RPC executor will not simulate to estimate or inject missing limits. Explicit resource-limit instructions already present in a transaction message are preserved. This can be useful for transactions that are close to the message size limit, where adding a compute-budget instruction would make an otherwise valid transaction too large.
 
 ## Deprecated plugins
 
--- a/packages/kit-plugin-rpc/src/resource-limit-estimation.ts
+++ b/packages/kit-plugin-rpc/src/resource-limit-estimation.ts
@@ -1,0 +1,9 @@
+/**
+ * Controls whether the RPC transaction planner and executor automatically
+ * reserve, estimate, and set transaction resource limits.
+ */
+export type ResourceLimitEstimationMode = 'estimate' | 'none';
+
+export function shouldEstimateResourceLimits(mode: ResourceLimitEstimationMode | undefined): boolean {
+    return mode !== 'none';
+}
--- a/packages/kit-plugin-rpc/src/solana-rpc.ts
+++ b/packages/kit-plugin-rpc/src/solana-rpc.ts
@@ -18,6 +18,7 @@ import { planAndSendTransactions } from '@solana/kit-plugin-instruction-plan';
 
 import { rpcAirdrop } from './airdrop';
 import { rpcGetMinimumBalance } from './get-minimum-balance';
+import { type ResourceLimitEstimationMode } from './resource-limit-estimation';
 import { rpcSubscriptionsConnection } from './rpc';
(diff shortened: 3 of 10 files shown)


lorisleiva:
@trevor-cortex

packages/kit-plugin-rpc/src/resource-limit-estimation.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Controls whether the RPC transaction planner and executor automatically
+ * reserve, estimate, and set transaction resource limits.
+ */
+export type ResourceLimitEstimationMode = 'estimate' | 'none';
+
+export function shouldEstimateResourceLimits(mode: ResourceLimitEstimationMode | undefined): boolean {
```
trevor-cortex, line 7: Two small things on this file:

1. **Missing docblock.** Per `CLAUDE.md`, all exported symbols need JSDoc. The `ResourceLimitEstimationMode` type above has one, but `shouldEstimateResourceLimits` does not.
2. **Is this meant to be public API?** `index.ts` re-exports everything from this file via `export *`, so `shouldEstimateResourceLimits` ends up on the package's public surface. It reads like an internal helper — only used by the planner and executor in this package. If it isn't intended for external use, consider either: exporting only the type from `index.ts` (e.g. `export type { ResourceLimitEstimationMode } from './resource-limit-estimation';`), or splitting the helper into a separate non-exported module. Keeps the public API tighter and avoids committing to the helper's signature.

packages/kit-plugin-rpc/src/solana-rpc.ts
```diff
@@ -111,15 +120,25 @@ export type SolanaRpcConfig<TClusterUrl extends ClusterUrl = ClusterUrl> = Solan
  * @see {@link solanaLocalRpc}
  */
 export function solanaRpc<TClusterUrl extends ClusterUrl>(config: SolanaRpcConfig<TClusterUrl>) {
-    return <T extends ClientWithPayer>(client: T) =>
-        pipe(
+    return <T extends ClientWithPayer>(client: T) => {
+        const resourceLimitEstimation =
+            config.resourceLimitEstimation ?? config.transactionConfig?.resourceLimitEstimation;
```
trevor-cortex, line 125: `resourceLimitEstimation` now lives in two places on `SolanaRpcConfig`: top-level, and nested inside `transactionConfig` (via `TransactionPlannerConfig`). The precedence here is "top-level wins, nested is fallback", which is reasonable, but it's not documented anywhere the user would see it — the top-level JSDoc doesn't mention that `transactionConfig.resourceLimitEstimation` is also honored, and the nested field's JSDoc doesn't mention it can be overridden.

A couple of options:

- **Pick one location.** Since this flag affects both planner *and* executor (not just message construction), a top-level-only option is arguably a better fit than nesting it under `transactionConfig`. That said, `TransactionPlannerConfig` needs to expose it too so `rpcTransactionPlanner` can be used standalone — that's fine, but `SolanaRpcConfig` could then read *only* the top-level field and ignore `transactionConfig.resourceLimitEstimation`.
- **Or keep both and document the precedence** in the top-level JSDoc ("falls back to `transactionConfig.resourceLimitEstimation` when not set at the top level").

Either is fine, but the current state — silently reading from two places — is easy to trip over.

packages/kit-plugin-rpc/README.md
```diff
@@ -301,6 +304,8 @@ Setting `skipPreflight: true` changes the behavior:
 | Estimation fails    | Throw                            | Use consumed resources, skip preflight |
 | Explicit limits set | Run preflight                    | Skip preflight                         |
 
+Set `resourceLimitEstimation: 'none'` to opt out of automatic resource-limit handling. The RPC planner will not reserve provisional resource-limit instructions, and the RPC executor will not simulate to estimate or inject missing limits. Explicit resource-limit instructions already present in a transaction message are preserved. This can be useful for transactions that are close to the message size limit, where adding a compute-budget instruction would make an otherwise valid transaction too large.
```
trevor-cortex, line 307: One small clarification worth adding: when `resourceLimitEstimation: 'none'` and `skipPreflight` is left at its default `false`, the executor still runs preflight during `sendAndConfirmTransaction` — since `didSimulateToEstimate` stays false, `skipPreflight: false` is passed through. That matches the "Explicit limits set" row in the table above, but a reader arriving at this paragraph looking to avoid *all* simulation overhead (e.g. for size-sensitive transactions) might not connect the two.

Consider a sentence like: "Note that preflight is unaffected — if you also want to skip the preflight simulation, set `skipPreflight: true`."

trevor-cortex (commented):
## Summary

Adds a `resourceLimitEstimation: 'estimate' | 'none'` opt-out that flows from `solanaRpc` into both `rpcTransactionPlanner` and `rpcTransactionPlanExecutor`. When set to `'none'`, the planner skips `fillTransactionMessageProvisoryResourceLimits` and the executor skips the `estimateAndSetResourceLimits` step in the send pipe. Default behavior is preserved, and the README/tests are updated to cover the new option.

The wiring is straightforward and the tests cover the important cases (top-level flag, nested `transactionConfig` flag, planner default, executor with estimation disabled asserting no `simulateTransaction` call and no CU limit set).

## Key things to watch out for

- **New file `resource-limit-estimation.ts` is exported from the package root**, which means `shouldEstimateResourceLimits` becomes part of the public API. It reads like an internal helper — worth considering whether it should be exported at all (see inline).
- **Missing docblock on `shouldEstimateResourceLimits`.** Per `CLAUDE.md`, all exported symbols need JSDoc docblocks. The type has one; the function does not.
- **Two places to configure the same thing on `SolanaRpcConfig`** (top-level `resourceLimitEstimation` and `transactionConfig.resourceLimitEstimation`) with silent precedence. Worth either documenting the precedence in the top-level JSDoc or picking a single location (see inline).
- **Interaction with preflight** is worth being explicit about in the README paragraph: opting out of estimation does *not* opt out of preflight — when `resourceLimitEstimation: 'none'` and `skipPreflight: false`, the send call will still run preflight, which is arguably the *only* simulation in that path. The current wording is accurate but could set the wrong expectation for someone opting out because they want zero simulation overhead.

- Changeset is `patch`. This is a new opt-in feature (new public option + new exported type), which would normally be `minor`, but the package is pre-1.0 (0.12.1) and the repo guidance allows `patch` at that stage — so this is defensible either way. Worth a sanity check against recent precedent in this repo.
- The `bufferAndRecoverResourceLimits`/`estimateResourceLimits` factories are still constructed unconditionally in the executor even when estimation is disabled. Harmless (they're just closures) and not worth changing.
- Behavioral parity looks good: with `'none'`, the planner leaves the message without provisory limits, the executor pipe short-circuits the estimate step and returns `tx` unchanged, so `context.message` and the signed transaction reflect exactly what the caller passed in (plus blockhash lifetime). That matches the README claim about preserving explicit resource-limit instructions.

lorisleiva:
Thanks! I agree with most of Trevor's comment if you don't mind having a look at them.

I'd just add a quick question on the API decision for `resourceLimitEstimation: "estimate" | "none"`. Since this is essentially a way to turn off resource limit estimation, I wonder if a simple boolean wouldn't be cleaner here? Unless you have ideas for future values that could potentially be added to the mix here?

lorisleiva:
Closing in favour of https://github.com/anza-xyz/kit-plugins/pull/344 🫡

PR #300, opened 2026-07-02
Upgrade to Kit v7

mcintyre94:
This will enable us to add react hooks to plugins.

PR #301, opened 2026-07-02
Add React hooks subpath to `@solana/kit-plugin-instruction-plan`

mcintyre94:
Adds the first set of React hooks, to the `instruction-plan` plugin, as a /react subpath (similar to what we did with kit for /query and /swr). 

This includes 4 hooks: `usePlanTransaction`, `usePlanTransactions`, `useSendTransaction`, and `useSendTransactions`

All are simple wrappers of `useAction` and `useClientCapability`.

### DIFF
--- a/packages/kit-plugin-instruction-plan/package.json
+++ b/packages/kit-plugin-instruction-plan/package.json
@@ -20,6 +20,7 @@
     "files": [
         "./dist/types",
         "./dist/index.*",
+        "./dist/react/index.*",
         "./src/"
     ],
     "type": "commonjs",
@@ -33,27 +34,58 @@
     "types": "./dist/types/index.d.ts",
     "react-native": "./dist/index.react-native.mjs",
     "exports": {
-        "types": "./dist/types/index.d.ts",
-        "react-native": "./dist/index.react-native.mjs",
-        "browser": {
-            "import": "./dist/index.browser.mjs",
-            "require": "./dist/index.browser.cjs"
+        ".": {
+            "types": "./dist/types/index.d.ts",
+            "react-native": "./dist/index.react-native.mjs",
+            "browser": {
+                "import": "./dist/index.browser.mjs",
+                "require": "./dist/index.browser.cjs"
+            },
+            "node": {
+                "import": "./dist/index.node.mjs",
+                "require": "./dist/index.node.cjs"
+            }
         },
-        "node": {
-            "import": "./dist/index.node.mjs",
-            "require": "./dist/index.node.cjs"
+        "./react": {
+            "types": "./dist/types/react/index.d.ts",
+            "react-native": "./dist/react/index.react-native.mjs",
+            "browser": {
+                "import": "./dist/react/index.browser.mjs",
+                "require": "./dist/react/index.browser.cjs"
+            },
+            "node": {
+                "import": "./dist/react/index.node.mjs",
+                "require": "./dist/react/index.node.cjs"
+            }
         }
     },
     "scripts": {
         "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
         "dev": "vitest --project node",
         "test": "pnpm test:types && pnpm test:treeshakability && pnpm test:unit",
-        "test:treeshakability": "for file in dist/index.*.mjs; do agadoo $file; done",
+        "test:treeshakability": "for file in dist/index.*.mjs dist/react/index.*.mjs; do agadoo $file; done",
         "test:types": "tsc --noEmit",
         "test:unit": "vitest run"
     },
+    "devDependencies": {
+        "@solana/react": "^7.0.0",
+        "@testing-library/react": "^16.0.0",
+        "@types/react": "^18.0.0 || ^19.0.0",
+        "react": "^18.0.0 || ^19.0.0",
+        "react-dom": "^18.0.0 || ^19.0.0"
+    },
     "peerDependencies": {
-        "@solana/kit": "^7.0.0"
+        "@solana/kit": "^7.0.0",
+        "@solana/react": "^7.0.0",
+        "react": ">=18"
+    },
+    "peerDependenciesMeta": {
+        "@solana/react": {
+            "optional": true
+        },
+        "react": {
+            "optional": true
+        }
     },
     "browserslist": [
         "supports bigint and not dead",
--- a/packages/kit-plugin-instruction-plan/package.json
+++ b/packages/kit-plugin-instruction-plan/package.json
@@ -53,7 +53,7 @@
         "test:unit": "vitest run"
     },
     "peerDependencies": {
-        "@solana/kit": "^6.10.0"
+        "@solana/kit": "^7.0.0"
     },
     "browserslist": [
         "supports bigint and not dead",
--- a/packages/kit-plugin-instruction-plan/src/react/__typetests__/index.ts
+++ b/packages/kit-plugin-instruction-plan/src/react/__typetests__/index.ts
@@ -1,0 +1,46 @@
+import type { ClientWithTransactionPlanning, ClientWithTransactionSending } from '@solana/kit';
+
+import { usePlanTransaction, usePlanTransactions, useSendTransaction, useSendTransactions } from '../index';
+
+// NB: hooks are only type-checked here, never executed.
+
+// [DESCRIBE] useSendTransaction
+{
+    const action = useSendTransaction();
+    // dispatch takes exactly the send input; data matches the client method's resolved value.
+    type Input = Parameters<ClientWithTransactionSending['sendTransaction']>[0];
+    type Result = Awaited<ReturnType<ClientWithTransactionSending['sendTransaction']>>;
+    action.dispatch(null as unknown as Input);
+    action.data satisfies Result | undefined;
+    // @ts-expect-error dispatch requires the input argument.
+    action.dispatch();
+    // @ts-expect-error dispatch takes only the input; there is no per-call config (the hook owns the abort signal).
+    action.dispatch(null as unknown as Input, {});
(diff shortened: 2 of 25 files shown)


mcintyre94:
* **#302** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#301** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#300** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-instruction-plan/test/useSendTransaction.react.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+        const returned = await hook.current.dispatchAsync(input);
+
+        expect(returned).toBe(result);
+        expect(sendTransactions).toHaveBeenCalledWith(
+            input,
+            expect.objectContaining({ abortSignal: expect.any(AbortSignal) }),
+        );
+    });
+});
```
trevor-cortex, line 76: The hook advertises that dispatching a second time aborts the in-flight call (see the docblocks in `src/react/index.ts` and the README section), but nothing in the two test files exercises that. Consider a test that dispatches, captures the `abortSignal` passed to the mock, dispatches again, and asserts the first signal is now aborted — otherwise the abort-on-redispatch contract is only enforced by whatever `@solana/react` does internally.

Same comment applies to `usePlanTransaction.react.test.tsx`.

packages/kit-plugin-instruction-plan/vitest.config.mts
```diff
@@ -1,9 +1,15 @@
...
+
 export default defineConfig({
     test: {
-        projects: [getVitestConfig('browser'), getVitestConfig('node'), getVitestConfig('react-native')],
+        projects: [
+            getVitestConfig('browser'),
+            mergeConfig(getVitestConfig('node'), excludeReactTests),
+            mergeConfig(getVitestConfig('react-native'), excludeReactTests),
+        ],
     },
```
trevor-cortex, line 14: The `/react` entry ships a `react-native` build, and this config runs the hook tests only in the browser project. Since the RN vitest project already uses happy-dom (see `vitest.config.base.mts`), including the react tests in RN would give a light smoke signal that the RN build path at least mounts. If there's a concrete reason to exclude it (e.g. a known incompatibility), a short comment here would help future readers; otherwise consider dropping the RN exclusion.

trevor-cortex (commented):
## Summary

Adds a `@solana/kit-plugin-instruction-plan/react` subpath exposing four thin hook wrappers (`usePlanTransaction`, `usePlanTransactions`, `useSendTransaction`, `useSendTransactions`) around the client's planning/sending methods, composed from `useClientCapability` + `useAction` in `@solana/react`. The hook manages the `AbortSignal` internally so a fresh dispatch aborts any in-flight call, and `dispatch`/`dispatchAsync` take just the input.

Build-wise, this is the repo's first multi-entry package. `tsup.config.base.ts` grows a backward-compatible optional `entry` param (default `['./src/index.ts']`), and only `kit-plugin-instruction-plan/tsup.config.ts` opts in — all other packages keep their existing single-entry build. The `/react` entry ships the full node/browser/react-native matrix, and the treeshakability script now covers it too.

`react` and `@solana/react` are declared as optional peer dependencies via `peerDependenciesMeta`, so the default entry point remains dependency-free and the hooks are strictly opt-in. Hook tests are `*.react.test.tsx` and only run in the browser (happy-dom) project; node and react-native projects `exclude` them via `mergeConfig`.

Overall the design is clean and the diff is well scoped. A few things worth a look:

## Things to watch

- **Abort-on-redispatch isn't test-covered.** The README, changeset, and every hook docblock advertise "an in-flight call is aborted when a newer one is dispatched", but no test exercises it — the `.react.test.tsx` files only assert that _an_ `AbortSignal` is threaded through, not that dispatching again aborts the previous signal. If this behavior comes entirely from `useAction` and is covered upstream in `@solana/react`, a one-line comment pointing at that would be enough; otherwise a redispatch test here would lock the contract down.
- **React-native project silently skips the hook tests.** `mergeConfig(getVitestConfig('react-native'), excludeReactTests)` excludes `**/*.react.test.tsx` from RN, but the `/react` entry _does_ ship a `react-native` build. Since the RN vitest project already uses happy-dom and `@testing-library/react` works fine there, dropping the RN exclusion would give a smoke signal that the RN build path renders. Not blocking — but worth deciding intentionally rather than by omission.
- **`useSendTransaction` return type may narrow away failure modes.** `sendTransaction`'s resolved value is (per the plugin) a `TransactionPlanResult` that can include partial-success cases. The type test derives `Awaited<ReturnType<...>>` directly, so whatever the client returns is what `data` is typed as — good, but worth eyeballing that the docblock's "single successful transaction" phrasing doesn't over-promise vs. the actual type.

- **Verify `useAction`'s callback signature.** All four hooks assume `useAction((abortSignal, input) => …)` — that `abortSignal` is the first argument and any additional args come from `dispatch(...)`. This is load-bearing for the whole design; worth a sanity check against the current `@solana/react` API if you have it handy.
- **`ActionResult<[input], Awaited<...>>` shape.** The type tests assert `dispatch()` requires the input and rejects a second argument. That contract lives entirely in `@solana/react`'s `ActionResult` type — if that ever changes to accept per-call options, the `@ts-expect-error` here will break loudly, which is the right behavior.
- **Umbrella package (`@solana/kit-plugins`)** is deprecated per CLAUDE.md and re-exports via `export *`, so subpath hooks won't leak through it — that's correct and matches the guidance to not add new exports to the umbrella.
- **Changeset bump level** (`minor`) matches repo policy for a new feature. No breaking changes to the default entry.
- The `test:treeshakability` glob now iterates `dist/index.*.mjs dist/react/index.*.mjs`. If a future refactor ever drops one of those directories, bash will pass the literal pattern to `agadoo` and fail with a slightly confusing error — non-issue today, just noting it.

packages/kit-plugin-instruction-plan/src/react/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    });
+    return useAction((abortSignal, input: SendTransactionInput) => client.sendTransaction(input, { abortSignal }));
+}
+
+type PlanTransaction = ClientWithTransactionPlanning['planTransaction'];
+type PlanTransactionInput = Parameters<PlanTransaction>[0];
+
+type PlanTransactions = ClientWithTransactionPlanning['planTransactions'];
+type PlanTransactionsInput = Parameters<PlanTransactions>[0];
+
```
trevor-cortex, line 43: Nit on grouping: the `PlanTransaction` / `PlanTransactions` type aliases sit here between `useSendTransaction` and `usePlanTransaction`, and then `SendTransactions` aliases appear separately near line 100. Not a correctness issue — everything is private to the module — but co-locating each alias pair with its hook (or grouping all four aliases at the top of the file) would make the file easier to scan.

packages/kit-plugin-instruction-plan/package.json
```diff
@@ -33,27 +34,58 @@
...
     "peerDependencies": {
-        "@solana/kit": "^7.0.0"
+        "@solana/kit": "^7.0.0",
+        "@solana/react": "^7.0.0",
+        "react": ">=18"
+    },
+    "peerDependenciesMeta": {
+        "@solana/react": {
+            "optional": true
+        },
```
trevor-cortex, line 85: Minor consistency thought: `react` is declared as `>=18` in `peerDependencies` but `^18.0.0 || ^19.0.0` in `devDependencies` / `@types/react`. `>=18` will happily accept a future React 20, which the maintainers may or may not want to promise support for from this changeset alone. Aligning on `^18.0.0 || ^19.0.0` in the peer range (matching the devDep floor) is the more conservative choice and matches the pattern used by many other Solana/React packages. Not blocking — just flagging in case there's a repo-wide convention here.

packages/kit-plugin-instruction-plan/src/react/__typetests__/index.ts
```diff
@@ -0,0 +1,46 @@
...
+{
+    const action = useSendTransaction();
+    // dispatch takes exactly the send input; data matches the client method's resolved value.
+    type Input = Parameters<ClientWithTransactionSending['sendTransaction']>[0];
+    type Result = Awaited<ReturnType<ClientWithTransactionSending['sendTransaction']>>;
+    action.dispatch(null as unknown as Input);
+    action.data satisfies Result | undefined;
+    // @ts-expect-error dispatch requires the input argument.
+    action.dispatch();
+    // @ts-expect-error dispatch takes only the input; there is no per-call config (the hook owns the abort signal).
```
trevor-cortex, line 17: Consider adding a positive assertion that `action.dispatch(null as unknown as Input)` returns `void` (or whatever `ActionResult` specifies) and that `action.dispatchAsync(input)` returns `Promise<Result>`. Right now the type test locks down the input side of the contract but doesn't assert the return side of `dispatch`/`dispatchAsync` — the `data satisfies Result | undefined` line covers the stored state but not the sync/async return types.

mcintyre94:
Closing - superseded by Kit PR: https://github.com/anza-xyz/kit/pull/1821

PR #302, opened 2026-07-03
Add React hooks subpath to `@solana/kit-plugin-wallet`

mcintyre94:
This PR adds react hooks to the wallet plugin.

Includes three hooks that expose the wallet state using `useSyncExternalStore`, each exposing one of the three pieces of state. `useWalletStatus`, `useConnectedWallet`, and `useWallets`. These are separate hooks because they change independently and will have different re-render cadence. You can still use `useSyncExternalStore(client.wallet.subscribe, client.wallet.getState)` to subscribe to the whole state, but there isn't a specific hook for this.

Also includes four new action hooks, which are all just simple wrappers around wallet plugin functions: `useConnect`, `useDisconnect`, `useSignIn` and `useSignMessage`. A hook is also provided for `useSelectAccount`, which is just a synchronous callback.

### DIFF
--- a/.changeset/sparkly-forks-wish.md
+++ b/.changeset/sparkly-forks-wish.md
@@ -1,0 +1,25 @@
+---
+'@solana/kit-plugin-wallet': minor
+---
+
+Add a `@solana/kit-plugin-wallet/react` subpath exposing React hooks for wallet state and actions. State hooks (`useWalletStatus`, `useConnectedWallet`, `useWallets`) subscribe to individual slices of `WalletState` via `useSyncExternalStore`; action hooks (`useConnect`, `useDisconnect`, `useSignIn`, `useSignMessage`, `useSelectAccount`) wrap the wallet actions. `react` and `@solana/react` are optional peer dependencies, so the default entry point is unchanged.
+
+```tsx
+import { useConnect, useConnectedWallet, useWallets, useWalletStatus } from '@solana/kit-plugin-wallet/react';
+
+function WalletButton() {
+    const status = useWalletStatus();
+    const wallets = useWallets();
+    const connected = useConnectedWallet();
+    const { dispatch: connect, isRunning } = useConnect();
+
+    if (status === 'pending') return null;
+    if (connected) return <p>Connected: {connected.account.address}</p>;
+
+    return wallets.map(wallet => (
+        <button key={wallet.name} disabled={isRunning} onClick={() => connect(wallet)}>
+            Connect {wallet.name}
+        </button>
+    ));
+}
+```
--- a/packages/kit-plugin-wallet/README.md
+++ b/packages/kit-plugin-wallet/README.md
@@ -169,11 +169,54 @@ All wallet state is accessed via `client.wallet.getState()`, which returns a ref
     const output = await client.wallet.signIn(selectedWallet, { domain: window.location.host });
     ```
 
+## React hooks
+
+The `@solana/kit-plugin-wallet/react` subpath exposes hooks for reading wallet state and driving wallet actions from React components. Install the optional peer dependencies (`react` and [`@solana/react`](https://www.npmjs.com/package/@solana/react)) and render your tree inside a `ClientProvider` whose client has a wallet plugin installed.
+
+The **state** hooks subscribe via `useSyncExternalStore`, each to a single slice of `WalletState`, so a component only re-renders when the slice it reads changes:
+
+| Hook                 | Returns                                                        |
+| -------------------- | -------------------------------------------------------------- |
+| `useWalletStatus`    | The current `WalletStatus`.                                    |
+| `useConnectedWallet` | The active connection (`{ account, signer, wallet }`) or null. |
+| `useWallets`         | The discovered wallets for the client's chain.                 |
+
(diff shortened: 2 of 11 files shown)


mcintyre94:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/302?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#302** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/302?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#301** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/301?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#300** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/300?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Adds a `@solana/kit-plugin-wallet/react` subpath exposing three state hooks (`useWalletStatus`, `useConnectedWallet`, `useWallets`) built on `useSyncExternalStore`, four action hooks (`useConnect`, `useDisconnect`, `useSignIn`, `useSignMessage`) that wrap `useAction`/`useClientCapability` from `@solana/react`, and a synchronous `useSelectAccount`. `react` and `@solana/react` are added as optional peer deps so the default entry point is unaffected. The package.json `exports` map is restructured from a bare shape to a nested `"."` + `"./react"` shape, tsup gets a second entry, declarations include the new file, and the vitest config excludes `*.react.test.tsx` from the node and react-native projects (they need happy-dom). This is the repo's first `/react` subpath, so the pattern established here will likely be copied to other plugins next.

The core implementation is very good — hooks are thin, uniformly structured through a shared `useWalletNamespace` helper, docblocks are complete with examples and `{@link}`s, the README additions are clear, and the tests cover both the state slice-subscription behavior and the abort-signal threading for actions. Nice touch that the type tests assert both success shapes and `@ts-expect-error` on missing arguments.

## ⚠️ Please remove before merge — four files that don't belong in this PR

These look accidental:

1. **`fable-review-comments.md`** at the repo root. This is a prior AI-generated code review of a *different* PR (kit-plugin-wallet store/types/tests, PR #191) with checkbox-tracked fixes. It's not the kind of artifact you want landing in the repo history.

2. **`docs/superpowers/plans/2026-07-02-instruction-plan-react-subpath.md`** and **`docs/superpowers/specs/2026-07-02-instruction-plan-react-subpath-design.md`** — these are the implementation plan and design spec for `@solana/kit-plugin-instruction-plan/react` (a different, future package), not for this wallet subpath. The plan file even self-instructs: "Do NOT commit the design spec or this plan file."

3. **`docs/superpowers/specs/2026-06-24-rpc-extension-plugins-design.md`** — a design spec for RPC extension plugins, also unrelated to this PR.

If there's a broader decision to start tracking agent working-docs in the repo, that should be a separate PR with a discussion; either way they shouldn't ride in with an unrelated feature.

## Notes for reviewers / follow-up

- **Exports restructure is a subtle break to double-check.** The old `exports` used bare top-level condition keys (`types`, `browser`, `node`, `react-native`). The new shape nests them under `"."` and adds `"./react"`. Node's resolver treats these differently — deep imports like `@solana/kit-plugin-wallet/dist/...` that worked before are now blocked (which is desired), but it's worth confirming the top-level condition ordering (`types` first, then `react-native`, then `browser`/`node`) still resolves the same way it did. The sibling `kit-plugin-signer` still uses the bare shape, so if this pattern gets copied later, worth a follow-up to align them.
- **`react` peer range is `>=18`.** No upper bound means a future React 20 would silently satisfy it. `@solana/react` uses `^7.0.0`, which is stricter. Matching ecosystem conventions might be worth checking, but the intent (broad compatibility as long as `useSyncExternalStore` exists) is clear.
- **The `test:treeshakability` script** now globs `dist/react/index.*.mjs` — verify locally that all five build targets (node cjs+esm, browser cjs+esm, react-native esm) actually pass `agadoo`, particularly the react-native one since these hooks pull in `react`.
- **`useSelectAccount` reference stability** is tested only across a rerender with the same wrapper. If a consumer swaps the `ClientProvider` value (i.e. reads a different `wallet` namespace), the returned function will change — that's correct behavior, just worth noting since the docblock says the reference is stable.
- Type test at `src/react/__typetests__/index.ts:26` uses `connected satisfies { account: UiWalletAccount } | null`. Since `WalletState['connected']` is nullable this holds, but if the connected shape gains more required fields, this assertion will still pass (it only checks `account`). The prior line's full-shape check covers it, so this is more a "be aware".

The implementation itself is ready to merge once the four stray files are removed.

PR #306, opened 2026-07-06
Add React hooks to `@solana/kit-plugin-wallet`

mcintyre94:
This PR adds react hooks to the wallet plugin.

Includes three hooks that expose the wallet state using `useSyncExternalStore`, each exposing one of the three pieces of state. `useWalletStatus`, `useConnectedWallet`, and `useWallets`. These are separate hooks because they change independently and will have different re-render cadence. You can still use `useSyncExternalStore(client.wallet.subscribe, client.wallet.getState)` to subscribe to the whole state, but there isn't a specific hook for this.

Also includes four new action hooks, which are all just simple wrappers around wallet plugin functions: `useConnect`, `useDisconnect`, `useSignIn` and `useSignMessage`. A hook is also provided for `useSelectAccount`, which is just a synchronous callback.

### DIFF
--- a/packages/kit-plugin-wallet/src/react/__typetests__/index.ts
+++ b/packages/kit-plugin-wallet/src/react/__typetests__/index.ts
@@ -1,0 +1,85 @@
+import type { ReadonlyUint8Array, SignatureBytes } from '@solana/kit';
+import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';
+import type { UiWallet, UiWalletAccount } from '@wallet-standard/ui';
+
+import type { WalletState, WalletStatus } from '../../types';
+import {
+    useConnect,
+    useConnectedWallet,
+    useDisconnect,
+    useSelectAccount,
+    useSignIn,
+    useSignMessage,
+    useWallets,
+    useWalletStatus,
+} from '../index';
+
+// NB: hooks are only type-checked here, never executed.
+
+// [DESCRIBE] useWalletStatus
+{
+    const status = useWalletStatus();
+    status satisfies WalletStatus;
+}
+
+// [DESCRIBE] useConnectedWallet
+{
+    const connected = useConnectedWallet();
+    connected satisfies WalletState['connected'];
+    // The connection is nullable when disconnected.
+    connected satisfies { account: UiWalletAccount } | null;
+}
+
+// [DESCRIBE] useWallets
+{
+    const wallets = useWallets();
+    wallets satisfies readonly UiWallet[];
+}
+
+// [DESCRIBE] useConnect
+{
+    const action = useConnect();
+    action.dispatch(null as unknown as UiWallet);
+    action.data satisfies readonly UiWalletAccount[] | undefined;
+    // @ts-expect-error dispatch requires the wallet argument.
+    action.dispatch();
+}
+
+// [DESCRIBE] useDisconnect
+{
+    const action = useDisconnect();
+    // The wallet argument is optional — omit it to disconnect the active wallet.
+    action.dispatch();
+    // Or pass a specific wallet to deauthorize.
+    action.dispatch(null as unknown as UiWallet);
+    action.data satisfies void | undefined;
+}
+
+// [DESCRIBE] useSignIn
+{
+    const action = useSignIn();
+    action.dispatch(null as unknown as UiWallet, null as unknown as SolanaSignInInput);
+    action.data satisfies SolanaSignInOutput | undefined;
+    // @ts-expect-error dispatch requires both the wallet and the sign-in input.
+    action.dispatch(null as unknown as UiWallet);
+}
+
+// [DESCRIBE] useSignMessage
+{
+    const action = useSignMessage();
+    action.dispatch(new Uint8Array());
+    action.data satisfies SignatureBytes | undefined;
+}
+{
+    // dispatch accepts a ReadonlyUint8Array (e.g. codec output) without a cast.
+    const action = useSignMessage();
+    action.dispatch(null as unknown as ReadonlyUint8Array);
+}
+
+// [DESCRIBE] useSelectAccount
+{
+    const selectAccount = useSelectAccount();
+    selectAccount(null as unknown as UiWalletAccount) satisfies void;
+    // @ts-expect-error selectAccount requires the account argument.
+    selectAccount();
+}
--- a/packages/kit-plugin-wallet/src/react/index.ts
+++ b/packages/kit-plugin-wallet/src/react/index.ts
@@ -1,0 +1,226 @@
+import type { ReadonlyUint8Array, SignatureBytes } from '@solana/kit';
+import { type ActionResult, useAction, useClientCapability } from '@solana/react';
+import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';
+import type { UiWallet, UiWalletAccount } from '@wallet-standard/ui';
+import { useSyncExternalStore } from 'react';
+
+import type { ClientWithWallet, WalletNamespace, WalletState, WalletStatus } from '../types';
+
+/**
+ * Reads the wallet namespace from the nearest `ClientProvider`, asserting at mount that a wallet
+ * plugin is installed. Shared by every hook in this module so the capability check, error hook
+ * name, and provider hint stay in one place.
+ */
+function useWalletNamespace(hookName: string): WalletNamespace {
+    return useClientCapability<ClientWithWallet>({
+        capability: 'wallet',
+        hookName,
+        providerHint: 'Install a wallet plugin on the client, e.g. `walletSigner()`.',
+    }).wallet;
+}
+
+// -- State hooks ------------------------------------------------------------
+
+/**
+ * Subscribes to the wallet connection status from a React component.
+ *
+ * Wraps `client.wallet.subscribe` / `getState` with `useSyncExternalStore`, re-rendering only when
+ * the status changes. Requires a `ClientProvider` whose client has a wallet plugin installed;
+ * asserts this at mount with {@link useClientCapability}.
+ *
+ * @returns The current {@link WalletStatus} (`'pending'` on the server and in React Native).
+ *
+ * @example
+ * ```tsx
+ * const status = useWalletStatus();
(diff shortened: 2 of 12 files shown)


mcintyre94:
* **#306** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/306?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/306?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/react/index.ts
```diff
@@ -0,0 +1,227 @@
...
+ * Wraps `client.wallet.disconnect` with {@link useAction}. Requires a `ClientProvider` whose client
+ * has a wallet plugin installed; asserts this at mount with {@link useClientCapability}.
+ *
+ * `dispatch`/`dispatchAsync` take an optional {@link UiWallet}. Called with no argument they
+ * disconnect the active wallet; passing a non-active, currently authorized wallet deauthorizes that
+ * wallet while leaving the active connection in place.
+ *
+ * @example
+ * ```tsx
+ * const { dispatch, isRunning } = useDisconnect();
```
trevor-cortex, line 137: Small nit on the abort semantics wording — this is shared with `useSignIn` below.

The observable "in-flight connect is aborted when a newer one is dispatched" behavior is real, but it doesn't actually come from the `AbortSignal`. Per `WalletActionOptions` in `types.ts`, the plugin only checks `abortSignal.throwIfAborted()` **pre-call** — once the wallet-standard `standard:connect` call has been dispatched, aborts via the signal do not take effect. What actually cancels the older call is the plugin's own supersession logic (throwing `DOMException` with `name: 'AbortError'` on the superseded `connect`/`signIn`).

End result for the user is the same, but the current wording implies the mechanism is the abort signal. Consider softening to something like:

> The hook owns an `AbortSignal` that is threaded into the wallet action; the plugin also treats a newer `connect`/`signIn` as superseding any in-flight one, so a stale dispatch rejects with `AbortError`.

Or just "an in-flight connect is superseded when a newer one is dispatched" without attributing it to the signal.

packages/kit-plugin-wallet/src/react/index.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @example
+ * ```tsx
+ * const { dispatch } = useSignMessage();
+ * dispatch(new TextEncoder().encode('Hello'));
+ * ```
+ */
+export function useSignMessage(): ActionResult<[message: ReadonlyUint8Array], SignatureBytes> {
+    const wallet = useWalletNamespace('useSignMessage');
+    return useAction((abortSignal, message: ReadonlyUint8Array) => wallet.signMessage(message, { abortSignal }));
+}
```
trevor-cortex, line 198: Same abort-semantics point as on `useConnect` — but more consequential here, because `signMessage` has **no** plugin-side supersession. Per `WalletActionOptions`, the signal is only honored pre-call, and unlike `connect`/`signIn`, `signMessage` doesn't document an `AbortError` from supersession.

So in practice: if a user dispatches a second `signMessage` while the first is waiting for the wallet, the first's abort signal fires — but if the wallet has already been dispatched, the first promise will still resolve with its signature (and the `ActionResult` for the older call will be discarded by `useAction`, which is fine). The wording "an in-flight signature request is aborted when a newer one is dispatched" is a bit stronger than what actually happens.

Suggest something like: "a stale in-flight dispatch is dropped from the hook's state when a newer one starts (though the wallet may still complete the older request in the background)."

packages/kit-plugin-wallet/src/react/__typetests__/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+// [DESCRIBE] useSignIn
+{
+    const action = useSignIn();
+    action.dispatch(null as unknown as UiWallet, null as unknown as SolanaSignInInput);
+    action.data satisfies SolanaSignInOutput | undefined;
+    // @ts-expect-error dispatch requires both the wallet and the sign-in input.
+    action.dispatch(null as unknown as UiWallet);
+}
+
```
trevor-cortex, line 66: Nit: `void | undefined` works, but `undefined` alone is tighter and matches the shape of `data` before dispatch more precisely. Feel free to ignore — the intent is clear either way.

packages/kit-plugin-wallet/src/react/index.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @example
+ * ```tsx
+ * const selectAccount = useSelectAccount();
+ * <button onClick={() => selectAccount(account)}>Use this account</button>
+ * ```
+ *
+ * @see {@link useConnectedWallet}
+ */
+export function useSelectAccount(): WalletNamespace['selectAccount'] {
+    return useWalletNamespace('useSelectAccount').selectAccount;
```
trevor-cortex, line 225: The docstring says "The returned reference is stable across renders." That's true in practice because the wallet namespace object built by the plugin is stable, so `namespace.selectAccount` is a stable method reference — but the stability is inherited from the plugin, not enforced by this hook (there's no `useCallback` / `useMemo` wrapper here). The test verifies it via a single `rerender()` with the same wrapper, which trivially exercises the plugin-side stability.

Worth clarifying either in the docstring ("…provided the underlying client is stable") or adding a `useCallback` here to make the hook itself the source of stability. Non-blocking — as written, callers relying on this for `useEffect` deps will Just Work with the current plugin, but the guarantee is a bit implicit.

trevor-cortex (commented):
## Summary

Adds a `@solana/kit-plugin-wallet/react` subpath with React hooks that wrap the existing wallet namespace:

- **State hooks** (`useWalletStatus`, `useConnectedWallet`, `useWallets`) subscribe to individual slices of `WalletState` via `useSyncExternalStore`. Splitting into slices is the right call — components only re-render on the slice they read, and the plugin's referentially-stable `getState()` contract makes the sliced `getSnapshot` closures safe.
- **Action hooks** (`useConnect`, `useDisconnect`, `useSignIn`, `useSignMessage`) wrap the async actions via `@solana/react`'s `useAction`, threading its `AbortSignal` into the plugin. `useSelectAccount` returns the sync callback directly.
- All hooks share `useWalletNamespace`, which asserts at mount that a wallet plugin is installed via `useClientCapability`.

Packaging is well done: `exports` map gains a `./react` entry with per-platform `types`/`browser`/`node`/`react-native` conditions, `react` and `@solana/react` are added as **optional** peer deps (default entry point unaffected), the `files` array picks up the new `dist/react/index.*` outputs, `tsconfig.declarations.json` includes the new entry, `tsup` builds it as a second entry via the extended `getPackageBuildConfigs({ entry })` helper in `tsup.config.base.ts`, and `test:treeshakability` now runs `agadoo` over `dist/react/index.*.mjs` too. The changeset is a proper `minor` and previews the API well.

Test coverage is solid: type tests cover argument/return shapes for every hook plus two `@ts-expect-error` assertions, and runtime tests cover both the mount-time capability assertion and the dispatch/abort-signal threading for actions plus subscribe/notify for state. Vitest is correctly configured to run the DOM tests only under the browser (happy-dom) project via a new `**/*.react.test.tsx` exclude on the node and react-native projects.

Overall this looks great to me — approving in spirit (I don't have push access on this repo so leaving it as a comment review). Just a few small wording and clarity nits below.

## Things to watch out for

- **Abort semantics wording.** A couple of docblocks say the hook "owns the `AbortSignal`, so an in-flight [action] is aborted when a newer one is dispatched." That's the observable behavior for `useConnect`/`useSignIn`, but the mechanism is the plugin's own supersession (throwing `AbortError` on the older call), not the signal — per `WalletActionOptions`, the signal is only honored **pre-call**. And for `useSignMessage` there is no plugin-side supersession at all, so a newer dispatch will *not* actually cancel an in-flight signature request. Details inline.
- **Optional peer deps + subpath.** The `./react` entry pulls `react` and `@solana/react` at import time. This is fine because non-React consumers never touch the subpath, but it's worth confirming downstream that only importing the root entry doesn't bundle these.

- The `ClientProvider` / `useAction` / `useClientCapability` primitives come from `@solana/react@^7.0.0`, which is added as a devDep and optional peer. Their contracts (dispatch signature, `ActionResult` shape, capability assertion) are exercised by the tests here rather than documented in this diff — worth a quick spot-check that the `^7.0.0` range is what actually ships those APIs.
- No integration test asserts that the state hooks correctly skip re-renders when *other* slices change (e.g. `useWalletStatus` should not re-render when only `wallets` changes). The current tests cover the positive path (updates propagate) but not the negative one. Non-blocking — the plugin's referential-stability contract makes this the plugin's responsibility, not the hooks'.
- Nothing platform-specific in the hooks themselves; the per-platform build variants exist for consistency with the rest of the package but the emitted JS should be identical across `browser`/`node`/`react-native`.

packages/kit-plugin-wallet/src/react/index.ts
mcintyre94: Updated to use suggested wording

packages/kit-plugin-wallet/src/react/index.ts
mcintyre94: Updated to use suggested wording

packages/kit-plugin-wallet/src/react/index.ts
mcintyre94: Removed this sentence, it's a property of the client/plugin not the hook.

lorisleiva (approved):
Very nice API!

mcintyre94:
### Merge activity

* **Jul 13, 2:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/306).
* **Jul 13, 2:31 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/306) rebased this pull request as part of a merge.
* **Jul 13, 2:31 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/306).

PR #314, opened 2026-07-10
wallet plugin: reflect updated authorized accounts in wallets list

mcintyre94:
This PR fixes a bug spotted while using the wallet plugin for our Kit react example app, which stress tests the multi wallet/multi account design. After connecting account(s) from a wallet, some parts of the UI would not reflect those changes.

### DIFF
--- a/packages/kit-plugin-wallet/test/store.test.ts
+++ b/packages/kit-plugin-wallet/test/store.test.ts
@@ -281,6 +281,30 @@ describe.skipIf(!__BROWSER__)('store (browser)', () => {
         expect(store.getState().connected!.account.address).toBe(newAccount.address);
     });
 
+    it('reflects accounts authorized during connect in the discovered-wallet list', async () => {
+        // Wallet starts with no authorized accounts (the realistic pre-connect state).
+        const mockWallet = createMockUiWallet({ accounts: [], name: 'TestWallet' });
+        registerWallet(mockWallet);
+
+        // Connect authorizes an account, regenerating the wallet's registry handle.
+        const account = createMockAccount();
+        connectMock.mockImplementationOnce(() => {
+            updateRegisteredWallet(createMockUiWallet({ accounts: [account], name: 'TestWallet' }));
+            return Promise.resolve();
+        });
+
+        const store = createWalletStore({ chain: 'solana:mainnet', storage: null });
+        await store.connect(mockWallet);
+
+        // `wallets` must reflect the newly authorized account immediately — not stay
+        // stale until an unrelated `standard:events change` fires. Consumers rendering
+        // the wallet from `useWallets()` rely on `.accounts` to know it is connected.
+        const state = store.getState();
+        expect(state.connected!.wallet.accounts.length).toBe(1);
+        expect(state.wallets[0].accounts.length).toBe(1);
+        expect(state.wallets[0]).toBe(state.connected!.wallet);
+    });
+
     it('transitions through connecting status', async () => {
         const account = createMockAccount();
         const mockWallet = createMockUiWallet({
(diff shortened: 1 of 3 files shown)


mcintyre94:
* **#314** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/314?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/314?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/test/store.test.ts
```diff
@@ -281,6 +281,29 @@ describe.skipIf(!__BROWSER__)('store (browser)', () => {
...
+        });
+
+        const store = createWalletStore({ chain: 'solana:mainnet', storage: null });
+        await store.connect(mockWallet);
+
+        // `wallets` must reflect the newly authorized account immediately — not stay
+        // stale until an unrelated `standard:events change` fires. Consumers rendering
+        // the wallet from `useWallets()` rely on `.accounts` to know it is connected.
+        const state = store.getState();
+        expect(state.connected!.wallet.accounts.length).toBe(1);
```
trevor-cortex, line 303: Minor: consider strengthening these two assertions to tie the list entry to the connected wallet directly, e.g.

```ts
expect(state.wallets[0]).toBe(state.connected!.wallet);
expect(state.wallets[0].accounts.length).toBe(1);
```

The current `.accounts.length` check passes as long as *some* handle for that wallet in `wallets` has an account, but the actual invariant the fix restores is that `wallets[i] === connected.wallet` after connect (i.e. the list holds the refreshed handle, not the stale pre-connect one). The reference-equality assertion pins that down and would also catch a future regression where the list is rebuilt from a different source than the ui-registry. Non-blocking.

trevor-cortex (commented):
Clean, well-scoped fix. `setConnected` is the single choke point for the three flows that authorize accounts (`connect`, `signIn`, silent reconnect), so reconciling `wallets` here closes the gap for all of them in one place. Using `reconcileWalletList()` (rather than `buildWalletList()` directly) keeps the snapshot referentially stable when the visible set is unchanged, which matches the existing pattern in `disconnectLocally` and the `change` handler. The changeset (patch), inline rationale, and test are all in good shape. LGTM.

**Notes for subsequent reviewers:**
- The core invariant is that after connect, the `wallets[i]` entry for the connected wallet holds the *refreshed* registry handle (with newly authorized accounts), not the pre-connect one. The test asserts `.accounts.length` — see the inline note for a stronger assertion that ties the list entry to `connected.wallet` directly.
- No regression risk for the no-op paths: `reconcileWalletList()` returns the existing reference on unchanged contents, so `updateState` won't emit a spurious snapshot when connect didn't add accounts (e.g. a wallet that pre-authorized).
- The symmetric fix applies to `signIn` and silent reconnect too, but they aren't independently tested. Since they all funnel through `setConnected` that's arguably fine — flagging for visibility rather than requesting more tests.

packages/kit-plugin-wallet/test/store.test.ts
mcintyre94: Updated this

Also added the symmetric tests for `signIn` and auto-reconnect mentioned

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Jul 13, 2:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/314).
* **Jul 13, 2:21 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/314).

PR #315, opened 2026-07-10
wallet plugin: subscribe to change events for all discovered wallets

mcintyre94:
Previously we only subscribed to change events for the connected wallet, which meant that if an authorized but not currently active wallet changed eg its account list then we wouldn't be subscribed to those changes.

This also enables adding cross-wallet account selection and disconnecting a non-active wallet, which are current API gaps.

### DIFF
--- a/.changeset/clean-candles-jog.md
+++ b/.changeset/clean-candles-jog.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-wallet': minor
+---
+
+Keep every discovered wallet's account state live by subscribing to wallet-standard `change` events for all discovered wallets, not just the connected one. This is an internal refactor with no public API change: `client.wallet.getState().wallets` now reflects account/feature changes for non-active wallets as they happen.
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -101,7 +101,11 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
 
     const listeners = new Set<() => void>();
     const signerListeners = new Set<() => void>();
-    let walletEventsCleanup: (() => void) | null = null;
+    // One `standard:events change` subscription per discovered wallet, keyed by
+    // name (the stable identity; the UiWallet handle regenerates on changes). The
+    // underlying subscription attaches to the raw wallet's events, so it survives
+    // handle regeneration — no need to resubscribe when a wallet's accounts change.
+    const walletEventSubscriptions = new Map<string, () => void>();
     let reconnectCleanup: (() => void) | null = null;
     let userHasSelected = false;
     let connectGeneration = 0;
@@ -154,6 +158,10 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
     function updateState(updates: Partial<WalletStoreState>): void {
         const prev = state;
         state = { ...state, ...updates };
+
+        if (state.wallets !== prev.wallets) {
+            syncWalletEventSubscriptions(state.wallets);
+        }
 
         // Signer subscribers (`subscribeToPayer` / `subscribeToIdentity`) fire
         // only when the observed signer changes.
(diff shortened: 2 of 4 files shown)


mcintyre94:
* **#315** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/315?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/315?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

Replaces the single-wallet `change` subscription with a per-wallet subscription managed in a `Map<name, cleanup>`, reconciled from `updateState` whenever `state.wallets` changes reference. Non-active wallets now stay live in `state.wallets` (accounts/features refresh as they happen), which unblocks cross-wallet account selection and disconnecting a non-active wallet down the line. Test setup is upgraded to match: `walletEventHandlers` is now keyed by wallet name, `emitWalletChange(name, props)` fires a specific one, and `getWalletForHandle` correctly resolves account handles via a `WeakMap` (mirroring the real registry).

LGTM overall — a couple of small observations below, none blocking.

## Design notes

- `syncWalletEventSubscriptions` is driven off reference equality of `state.wallets`, which is already carefully preserved by `reconcileWalletList` — so it correctly no-ops on register/change events that don't change the visible set, and correctly re-runs when it does. Filtered-out wallets are never subscribed, which is the right call (register/unregister flips eligibility).
- Keying subscriptions by `name` while the underlying subscription attaches to the raw wallet is the right identity choice: the UiWallet handle regenerates on every accounts/features change, but the raw wallet (and therefore the events feature) is stable, so we don't need to resubscribe on handle regeneration. The header comment on `walletEventSubscriptions` calls this out well.
- The `change` handler's active-wallet path preserves the previous reconcile logic verbatim; the non-active path is a tight `updateState({ wallets })`, and both share the single `reconcileWalletList()` call up top. Nice factoring.
- Dispose iterates the map and clears — correct, and `walletEventsCleanup` is fully gone. I also verified the two other former call sites (`disconnectLocally`, the unregister handler) no longer need to touch subscriptions: unregister flows through `syncWalletEventSubscriptions` via the wallets-reference change, and a plain disconnect intentionally leaves the wallet subscribed (it's still discovered).

## Things to watch out for

- **Doc regression on `subscribeToWalletEvents`**: the old block comment above `eventsFeature.on('change', ...)` explained *why* the handler ignores the payload flags and treats the account reference as the single source of truth for signer churn. That reasoning is still true and non-obvious; the new inline comments cover the *what* but not the *why*. Consider re-adding the deleted paragraph (or a condensed version) so the next reader doesn't wonder whether they should be branching on `properties.accounts` / `properties.features`.
- **Reentrancy**: `syncWalletEventSubscriptions` runs inside `updateState`, before listeners are notified. In practice `on('change')` never fires synchronously on subscribe (per wallet-standard), so there's no reentrancy in reality — but it's worth being aware of if a misbehaving wallet ever did. Not a blocker.
- **`walletEventHandler` back-compat in tests**: `_setup.ts` still exports the single-handler `walletEventHandler` alongside the new per-wallet map, with a "last subscriber wins" note. Since the store now subscribes to every discovered wallet, whichever handler "wins" depends on registration order — existing tests happen to work because they only register one wallet, but it's a subtle footgun for future test authors. Might be worth a follow-up to migrate the remaining uses to `emitWalletChange` and drop the back-compat export.

- The changeset is correctly marked `minor` and framed as an internal refactor with no public API change — matches what the PR actually does.
- New tests cover: non-active wallet keeps live accounts, late-registered wallets get subscribed, unregister tears down the subscription, dispose clears all, and no-op non-active changes don't notify listeners. Good coverage of the invariants that matter.
- Worth spot-checking: the change handler's active-wallet path when the connected wallet drops its chain/feature. `newWallets` is computed first (excluding the now-filtered wallet), then `filterWallet(refreshed)` fails, then `disconnectLocally()` runs and computes `reconcileWalletList()` a second time. Both produce the same list, so it's correct — just a redundant rebuild. Not worth changing for a hot-path that fires rarely.

mcintyre94:
> Doc regression on subscribeToWalletEvents

Updated

> walletEventHandler back-compat in tests: _setup.ts still exports the single-handler walletEventHandler alongside the new per-wallet map, with a "last subscriber wins" note.

Agreed, removed the back compat that was being used for tests and they now exercise `emitWalletChange`

lorisleiva (approved):

mcintyre94:
### Merge activity

* **Jul 13, 2:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/315).
* **Jul 13, 2:23 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/315) rebased this pull request as part of a merge.
* **Jul 13, 2:23 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/315).

PR #316, opened 2026-07-10
wallet plugin: cross-wallet selectAccount and targeted disconnect

mcintyre94:
`selectAccount(account)` now accepts an authorized account from any wallet, not only the currently active one. This enables the ability to switch between accounts in different wallets in the Kit example app, which the non-plugin implementation has. It is still synchronous, as it is only for already authorized accounts and doesn't need to communicate with the wallet.

`disconnect()` now has an optional first argument `wallet`, which allows disconnecting a wallet that is not currently active. This is a capability the react example app had, and means you don't need to switch to an account in wallet A to disconnect wallet A from the app. If the  `wallet` arg is omitted, or is the currently active wallet, then the existing behaviour is retained - the active wallet is disconnected, and the wallet state reflects no active connection.

This is a breaking change, as there was already an optional `options` argument which now needs to be the second arg.

### DIFF
--- a/.changeset/fair-kids-shine.md
+++ b/.changeset/fair-kids-shine.md
@@ -1,0 +1,13 @@
+---
+'@solana/kit-plugin-wallet': minor
+---
+
+Add cross-wallet account selection and targeted disconnect.
+
+`selectAccount` now accepts an account belonging to any authorized wallet, not just the active one. When the account belongs to a different authorized wallet, the active connection switches to it synchronously (no reconnect prompt), leaving the previously-active wallet authorized. The owning wallet is resolved from the account handle's identity, so selection is unambiguous even when the same address is authorized in two different wallets.
+
+This also means that we can synchronously transition from `disconnected` to `connected` by calling `selectAccount` with a previously authorized but not currently active account. This would previously throw `WALLET__NOT_CONNECTED`. Now throws `WALLET__ACCOUNT_NOT_AVAILABLE` only when the selected account is not available/not authorized.
+
+`disconnect` now takes an optional leading `wallet` argument — `disconnect(wallet?, options?)`. Passing a non-active authorized wallet deauthorizes just that wallet (calling `standard:disconnect` if supported) while leaving the active connection, status, and persisted account untouched; a wallet that is already gone or lacks `standard:disconnect` is a forgiving no-op. Calling `disconnect()` with no argument still disconnects the active wallet as before.
+
+**Breaking:** `client.wallet.disconnect` gained a leading `wallet` parameter. Callers passing options as the first argument must move them to the second: `disconnect(options)` becomes `disconnect(undefined, options)`.
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -477,17 +477,14 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
         }
     }
 
-    async function disconnect(options?: WalletActionOptions): Promise<void> {
+    async function disconnect(wallet?: UiWallet, options?: WalletActionOptions): Promise<void> {
         options?.abortSignal?.throwIfAborted();
 
-        // No connected wallet, but an auto-reconnect may still be in flight:
-        // during 'reconnecting' a silent reconnect or the late-registration
-        // listener can complete and connect, overriding this explicit
-        // disconnect. Record the user's intent so auto-connect won't re-arm,
-        // tear down the late-registration listener, and bump the generation so
-        // any in-flight connect/signIn/silent-reconnect bails at its guard
-        // instead of resuming.
-        if (!state.connectedWallet) {
+        const target = wallet ?? state.connectedWallet;
+
+        // Nothing to disconnect (no arg, nothing active): cancel any pending reconnect, and
+        // supersede in-flight attempts.
+        if (!target) {
             userHasSelected = true;
             cancelReconnect();
             connectGeneration++;
@@ -496,32 +493,63 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
             return;
         }
 
-        const currentWallet = state.connectedWallet;
-        // Bump the generation so this disconnect supersedes any connect/signIn
-        // that was already in flight (its prompt opened before the user chose to
-        // disconnect): that attempt captured an earlier generation, so it will
-        // bail at its guard rather than establishing a connection the user has
-        // since dismissed. Mirrors the not-connected branch above — newest action
-        // wins. A connect/signIn started *after* this disconnect bumps the
-        // generation again, so the `finally` guard skips `disconnectLocally` and
-        // leaves the newer connection intact.
-        const generation = ++connectGeneration;
-        // Mark this wallet as disconnecting
-        disconnectingWalletName = currentWallet.name;
-        updateState({ status: 'disconnecting' });
+        const isActive = state.connectedWallet != null && state.connectedWallet.name === target.name;
 
-        try {
-            if (currentWallet.features.includes(StandardDisconnect)) {
-                const disconnectFeature = getWalletFeature(
-                    currentWallet,
-                    StandardDisconnect,
-                ) as StandardDisconnectFeature[typeof StandardDisconnect];
-                await disconnectFeature.disconnect();
+        if (isActive) {
+            const currentWallet = state.connectedWallet!;
+            // Bump generation so this disconnect supersedes any in-flight connect/signIn.
+            const generation = ++connectGeneration;
+            disconnectingWalletName = currentWallet.name;
+            updateState({ status: 'disconnecting' });
+            try {
+                if (currentWallet.features.includes(StandardDisconnect)) {
+                    const disconnectFeature = getWalletFeature(
+                        currentWallet,
+                        StandardDisconnect,
+                    ) as StandardDisconnectFeature[typeof StandardDisconnect];
+                    await disconnectFeature.disconnect();
+                }
+            } finally {
+                if (generation === connectGeneration) {
+                    disconnectLocally();
+                }
             }
+            return;
+        }
+
+        // targeted disconnect but wallet gone: leave the active connection, status,
+        // and persisted key untouched.
+        if (!isWalletStillAvailable(target)) {
+            return; // already gone / not authorized — forgiving no-op
+        }
+        const refreshedTarget = refreshUiWallet(target);
+        if (!refreshedTarget.features.includes(StandardDisconnect)) {
+            return; // can't deauthorize without standard:disconnect — no-op
+        }
+        // Guard against a concurrent select/connect into this wallet for the await;
+        // restore the prior marker afterward (it may belong to an active disconnect).
+        const prevDisconnecting = disconnectingWalletName;
+        disconnectingWalletName = refreshedTarget.name;
+        try {
+            const disconnectFeature = getWalletFeature(
+                refreshedTarget,
+                StandardDisconnect,
+            ) as StandardDisconnectFeature[typeof StandardDisconnect];
+            await disconnectFeature.disconnect();
         } finally {
-            // Only clean up if no new connect/signIn started while we were awaiting.
-            if (generation === connectGeneration) {
-                disconnectLocally();
+            if (disconnectingWalletName === refreshedTarget.name) {
+                disconnectingWalletName = prevDisconnecting;
+            }
+            // Reflect the now-empty accounts immediately — but only if the store
+            // is still live. If it was disposed while `standard:disconnect`
+            // awaited, the disposer has already torn down every wallet-event
+            // subscription and cleared the map; deauthorizing the target
+            // regenerated its handle, so `reconcileWalletList` returns a fresh
+            // array, and letting `updateState` run here would re-invoke
+            // `syncWalletEventSubscriptions` and re-subscribe every wallet into
+            // the cleared map — listeners the disposer can no longer clean up.
+            if (!disposed) {
+                updateState({ wallets: reconcileWalletList() });
             }
         }
     }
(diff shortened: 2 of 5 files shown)


mcintyre94:
* **#316** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/316?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/316?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

.changeset/fair-kids-shine.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+Add cross-wallet account selection and targeted disconnect.
+
+`selectAccount` now accepts an account belonging to any authorized wallet, not just the active one. When the account belongs to a different authorized wallet, the active connection switches to it synchronously (no reconnect prompt), leaving the previously-active wallet authorized. The owning wallet is resolved from the account handle's identity, so selection is unambiguous even when the same address is authorized in two different wallets.
+
+`disconnect` now takes an optional leading `wallet` argument — `disconnect(wallet?, options?)`. Passing a non-active authorized wallet deauthorizes just that wallet (calling `standard:disconnect` if supported) while leaving the active connection, status, and persisted account untouched; a wallet that is already gone or lacks `standard:disconnect` is a forgiving no-op. Calling `disconnect()` with no argument still disconnects the active wallet as before.
+
+**Breaking:** `client.wallet.disconnect` gained a leading `wallet` parameter. Callers passing options as the first argument must move them to the second: `disconnect(options)` becomes `disconnect(undefined, options)`.
```
trevor-cortex, line 13: The changeset is thorough on `disconnect`'s signature change, but doesn't mention two other observable behaviour shifts introduced by the `selectAccount` change:

1. **Adopting a connection from `disconnected`.** With the `if (!state.connectedWallet) throw NOT_CONNECTED` guard gone, `selectAccount(authorizedAccount)` now synchronously transitions the store from `disconnected` → `connected` when the account belongs to a still-authorized (e.g. auto-connect-eligible) wallet. That's a useful new capability but it's a new side effect for existing callers.
2. **Error-code change for the same input.** `selectAccount` called while disconnected used to throw `WALLET__NOT_CONNECTED`; it now throws `WALLET__ACCOUNT_NOT_AVAILABLE` when the account's wallet isn't resolvable/authorized. Callers with `catch` blocks keyed on `NOT_CONNECTED` will silently stop matching.

Probably worth a sentence each in the changeset body. Neither warrants a `major` bump on its own (and pre-1.0 `minor` is already treated as breaking per the repo's changesets skill), but they're the kind of thing a user reading the changelog would want flagged.

trevor-cortex (commented):
## Summary

Extends `client.wallet` with two capabilities the non-plugin example app already had:

- `selectAccount(account)` now resolves the owning wallet from the account handle (via `getWalletForHandle`), so it can switch the active connection to an account of any currently authorized wallet — synchronously, no reconnect prompt. As a side effect, it can also *establish* a connection from `disconnected` when the account belongs to a still-authorized wallet.
- `disconnect()` now takes an optional leading `wallet` argument. Passing a non-active authorized wallet deauthorizes just that wallet (calls `standard:disconnect` if supported) without touching the active connection, status, or persisted account. `disconnect()` and `disconnect(activeWallet)` retain the prior behaviour.

Breaking: `disconnect`'s options argument moved from position 1 to position 2. Package is at `0.12.0` (pre-1.0), so `minor` is fine per the changesets skill.

The implementation is careful. In particular the non-active `disconnect` branch:

- Doesn't bump `connectGeneration` (so it doesn't interfere with active connect/signIn/disconnect flows).
- Temporarily claims `disconnectingWalletName` for the target across the await to block a concurrent `selectAccount` into it, and restores the prior marker only if it hasn't been overwritten by an active disconnect in the meantime — the interleavings I traced (concurrent active-disconnect, concurrent non-active-disconnect of same/different wallets) all resolve correctly.
- Adds a `!disposed` guard around the trailing `updateState`, since — unlike the active branch, which is protected by the `generation === connectGeneration` check — this branch would otherwise re-run `syncWalletEventSubscriptions` after `[Symbol.dispose]` has torn down and cleared the subscription map, leaking listeners the disposer can no longer clean up. The dispose-mid-await test covers this well.

A few things worth addressing before merge — nothing structural, mostly docs / a couple of test-coverage gaps. Details inline and below.

## Points that don't tie cleanly to a changed line

### README not updated (`packages/kit-plugin-wallet/README.md`)

CLAUDE.md's checklist calls out: *"When modifying an existing API, keep the README in sync."* Two items under the "State and actions" section:

- The **`disconnect()`** bullet reads "Disconnect the active wallet." — should mention the optional `wallet` argument and the targeted-disconnect semantics.
- The **`selectAccount(account)`** bullet reads "Switch to a different account within an already-authorized wallet without reconnecting." — technically still true, but doesn't convey that the account can belong to *any* currently authorized wallet (not just the connected one), and that this can also adopt a connection from `disconnected`.

Both are small textual updates; happy to defer to a follow-up if you'd rather keep this PR focused, but per the repo checklist they should ship together with the API change.

### Test-coverage gaps in the new `disconnect targeting` suite

Two cases worth adding while the harness is fresh (both in `packages/kit-plugin-wallet/test/store.test.ts`, in the new `store disconnect targeting (browser)` describe):

1. **Abort signal on the non-active branch.** The existing abort test in the `store action abort signals` describe only exercises `disconnect(undefined, { abortSignal })`. Adding a `disconnect(nonActiveWallet, { abortSignal: alreadyAbortedSignal })` case would pin down that the pre-await `throwIfAborted()` fires and no side effects occur (no `disconnectMock` call, no `disconnectingWalletName` mutation observable via a subsequent `selectAccount`).
2. **Non-active disconnect does not cancel a pending reconnect.** The active branch's no-target path calls `cancelReconnect()`; the non-active path deliberately does not. A test that arms the reconnect-wait path (`status: 'reconnecting'`) and then calls `disconnect(someOtherAuthorizedWallet)` and asserts status stays `'reconnecting'` would lock this in — right now nothing prevents a future refactor from "tidying up" and calling `cancelReconnect()` here, silently breaking that guarantee.

### `selectAccount` while owner is mid-disconnect

The `owner.name === disconnectingWalletName` branch in the new `selectAccount` is the only path that still throws `NOT_CONNECTED` instead of `ACCOUNT_NOT_AVAILABLE`, and it's now checking the *derived owner* rather than `state.connectedWallet` — a subtle meaning shift from the old code. Worth an explicit test that: connect wallet A, start `disconnect(A)` but don't await, then `selectAccount(aAccount)` from A, and assert `NOT_CONNECTED`. Same for the cross-wallet case: connect A, start `disconnect(B)` (non-active) with a hanging mock, then `selectAccount(bAccount)` and assert `NOT_CONNECTED`.

- **Verify the `disconnectingWalletName` interleavings.** The save/restore-if-unchanged pattern in the non-active branch is subtle; worth another set of eyes on the case where an active `disconnect(active)` starts *while* a non-active `disconnect(other)` is awaiting. My read: the active branch clears the marker via `disconnectLocally`, the non-active `finally` sees a mismatched marker and doesn't restore, ends up null — which is correct. But it's the kind of thing that benefits from a second reader.
- **Silent error-code shift on `selectAccount`.** Calling `selectAccount(someAccount)` while disconnected used to throw `WALLET__NOT_CONNECTED`; it now throws `WALLET__ACCOUNT_NOT_AVAILABLE` if the account's wallet isn't discoverable, or *succeeds* (adopts the connection) if it is. Not a bug — arguably an improvement — but not called out in the changeset (see inline). Users catching `NOT_CONNECTED` on `selectAccount` will see the error class change.

packages/kit-plugin-wallet/src/types.ts
```diff
@@ -263,13 +269,18 @@ export type WalletNamespace = {
...
+     * previously active wallet authorized.
      *
-     * @throws `SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED)` if no wallet is
-     *   connected, or the connected wallet is currently disconnecting.
      * @throws `SolanaError(SOLANA_ERROR__WALLET__ACCOUNT_NOT_AVAILABLE)` if the
-     *   specified account is not among the connected wallet's accounts.
+     *   account's handle cannot be resolved to a currently authorized wallet, or
+     *   the account is not among that wallet's accounts.
+     * @throws `SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED)` if the account's
+     *   wallet is currently disconnecting.
```
trevor-cortex, line 284: The updated docblock covers the cross-wallet switch case well, but doesn't mention that `selectAccount` can now be called while the store is `disconnected` to *adopt* an authorized wallet's account (synchronously, no prompt) — that's a new capability, and the tests exercise it explicitly (`adopts a connection when selecting an authorized account while disconnected`). Worth one sentence, e.g. after the existing paragraph:

> If the store is currently `'disconnected'`, selecting an account belonging to a still-authorized wallet establishes the active connection to that wallet synchronously — no reconnect prompt.

Also, the `@throws NOT_CONNECTED` line now only covers the mid-disconnect case; a reader who remembers the old behaviour may wonder what happened to the disconnected case. Either the changeset's error-code note (see comment there) or a `@remarks` here would help.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -488,32 +485,63 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
...
+            // is still live. If it was disposed while `standard:disconnect`
+            // awaited, the disposer has already torn down every wallet-event
+            // subscription and cleared the map; deauthorizing the target
+            // regenerated its handle, so `reconcileWalletList` returns a fresh
+            // array, and letting `updateState` run here would re-invoke
+            // `syncWalletEventSubscriptions` and re-subscribe every wallet into
+            // the cleared map — listeners the disposer can no longer clean up.
+            if (!disposed) {
+                updateState({ wallets: reconcileWalletList() });
             }
```
trevor-cortex, line 553: Two small notes on the non-active disconnect branch:

1. **Forgiving no-ops emit no state update.** The `!isWalletStillAvailable(target)` and `!refreshedTarget.features.includes(StandardDisconnect)` early returns skip the trailing `updateState({ wallets: reconcileWalletList() })`. Nothing's broken (active state hasn't changed), but a caller expecting `disconnect(nonActive)` to always at least surface the latest wallet-list snapshot won't get one on these paths. Given the intent ("forgiving no-op") I think that's the right call — worth a one-liner in the `types.ts` docblock clarifying no state is emitted in these cases, so callers relying on the returned Promise's resolution to know "the list is now current" aren't misled.
2. **Stylistic:** consider `if (disposed) return;` immediately after the `await disconnectFeature.disconnect();` to mirror the pattern in `setConnected`. The `finally` guard already prevents the actual re-subscription bug, so this is purely for symmetry — feel free to ignore.

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -558,31 +586,55 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
...
             updateState({ status: 'connected' });
             return;
         }
+
+        // Switch the active connection (owner may differ from the current one). The
+        // previously-active wallet is left authorized — we never disconnect it. Its
+        // event subscription persists, so switching back stays live.
         const signer = tryCreateSigner(selectedAccount);
         updateState({ account: selectedAccount, connectedWallet: refreshed, signer, status: 'connected' });
         persistAccount(selectedAccount, refreshed);
```
trevor-cortex, line 648: The handle-based owner resolution is a clean fix for the same-address-in-two-wallets ambiguity — nice call using `getWalletForHandle`'s WeakMap semantics rather than address matching.

One thing worth confirming: `getWalletForHandle(account)` throws for a stale/unregistered handle, which you correctly map to `ACCOUNT_NOT_AVAILABLE`. But `getOrCreateUiWalletForStandardWallet` on a stale-but-still-alive raw wallet returns *a fresh handle*, which then fails `isWalletStillAvailable(owner)` by name — also correctly mapped. So there are two doors to `ACCOUNT_NOT_AVAILABLE` here, both covered by tests. Just noting the double-guard is intentional and works, so a subsequent reader doesn't try to "simplify" one away.

mcintyre94:
> README not updated

Updated

> Test-coverage gaps in the new disconnect targeting suite

Agree, added these tests

> selectAccount while owner is mid-disconnect

Added a test for this case too

.changeset/fair-kids-shine.md
mcintyre94: Added a note on both

packages/kit-plugin-wallet/src/types.ts
mcintyre94: Updated this. I didn't add a remark about old behaviour because I think that'd be confusing, but it's mentioned in the changeset now.

packages/kit-plugin-wallet/src/store.ts
mcintyre94: - The no-op is intentional, it's when we can't disconnect the wallet, so there's no state update. Not sure extra documentation beyond no-op is necessary.
- I think the current formulation is a tiny bit clearer, but either works.

lorisleiva (approved):
Nice changes! Very nice that we can have multiple connected wallets at once. I wasn't sure about the `disconnect` API change versus adding a new function like `disconnect(wallet) + disconnectActive()` but I think having a single disconnect function is probably best here.

mcintyre94:
### Merge activity

* **Jul 13, 2:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/316).
* **Jul 13, 2:25 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/316) rebased this pull request as part of a merge.
* **Jul 13, 2:25 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/316).

PR #318, opened 2026-07-10
Widen wallet `signMessage` to accept a `ReadonlyUint8Array`

mcintyre94:
The wallet `signMessage` takes a `Uint8Array`, forcing callers with a Kit `ReadonlyUint8Array` to cast. This PR changes it to accept `ReadonlyUint8Array`.

The wallet-standard feature still uses `Uint8Array` so we cast at that boundary.

PR #317, opened 2026-07-10
Make wallet-plugin `payer`/`identity` getters non-enumerable

mcintyre94:
The dynamic `payer` and `identity` getters that `walletSigner`, `walletPayer`, and `walletIdentity` install on the client throw when no signer is connected. This is by design and allows us to keep the same signature for them. But they were configured as `enumerable: true`, which meant that code like object spread, `Object.entries`, and something deep in React's internals, would throw when they were unset.

Making them non-enumerable means that code like this doesn't throw, but detection like `'payer' in client`, and explicit access (which of course throws) still works.

PR #319, opened 2026-07-10
Add readiness helpers for the initial auto-reconnect warm-up

mcintyre94:
When the wallet plugin is first created, it has a `pending` state. If it runs auto-reconnect then it uses a `reconnecting` state for this. Then it settles to disconnected/connected. These are intentionally different states to the action-initiated connecting/disconnecting.

This PR adds helpers that allow apps to wait for the plugin to settle to a ready state before rendering it. This is useful to avoid flashing a 'reconnecting' state either on first load or when the Kit client changes, for example when changing chain.

- `client.wallet.whenReady()` returns a promise that resolves after the status settles. This is primarily for apps that are swapping the client, they can await this promise before rendering with the new client. This promise never rejects.
- `isWalletWarmingUp(status)` is a simple predicate, true if the status is `pending` or `reconnecting`, else false. Exported so apps don't need to worry about this plugin implementation detail if they just want a simple ready check for the current status.

In the new `/react` subpath:
- `useIsWalletReady()` is a hook to check if the wallet is ready 
- `<WalletReadyGate fallback={...}>{wallet UI}</WalletReadyGate>` is a React component that can be used to gate rendering UI until the wallet is ready. Useful for eg a wallet connect button, or anything rendering based on the connected state. 

Note that I plan to generalise the `client.wallet.whenReady` pattern into a composable plugin feature similar to dispose, so plugins can just register their `whenReady`, and apps can await a single `whenReady` promise at the client level without knowing which plugins implement it. That is primarily work in the Kit repo, and would just use `client.wallet.whenReady` as the wallet plugin's implementation. So all of this is still required - it'll just get easier to use for apps later.

### DIFF
--- a/packages/kit-plugin-wallet/src/react/index.ts
+++ b/packages/kit-plugin-wallet/src/react/index.ts
@@ -2,8 +2,10 @@ import type { ReadonlyUint8Array, SignatureBytes } from '@solana/kit';
 import { type ActionResult, useAction, useClientCapability } from '@solana/react';
 import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';
 import type { UiWallet, UiWalletAccount } from '@wallet-standard/ui';
+import type { ReactNode } from 'react';
 import { useSyncExternalStore } from 'react';
 
+import { isWalletWarmingUp } from '../status';
 import type { ClientWithWallet, WalletNamespace, WalletState, WalletStatus } from '../types';
 
 /**
@@ -94,6 +96,37 @@ export function useWallets(): readonly UiWallet[] {
     const wallet = useWalletNamespace('useWallets');
     const getWallets = () => wallet.getState().wallets;
     return useSyncExternalStore(wallet.subscribe, getWallets, getWallets);
+}
+
+/**
+ * Reports whether the wallet client has finished its initial auto-reconnect "warm-up".
+ *
+ * Returns `false` while the status is `'pending'` or `'reconnecting'` — the transient states a
+ * freshly built client passes through while it silently re-establishes a persisted connection — and
+ * `true` once it settles into any stable state (`'connected'`, `'disconnected'`, or a user-initiated
+ * `'connecting'` / `'disconnecting'`). This is the declarative, render-time counterpart to the
+ * store's imperative `whenReady()`, gating on the exact same set of statuses.
+ *
+ * Use it to hold back wallet-dependent UI until the connection has settled, so a persisted wallet
+ * never flashes "disconnected" before it reconnects. Re-renders only when this readiness boolean
+ * flips, not on every status change. Requires a `ClientProvider` whose client has a wallet plugin
+ * installed; asserts this at mount with {@link useClientCapability}.
+ *
+ * @returns `true` once the warm-up has settled, otherwise `false`.
+ *
+ * @example
+ * ```tsx
+ * const isReady = useIsWalletReady();
+ * return <button disabled={!isReady}>Connect</button>;
+ * ```
+ *
+ * @see {@link WalletReadyGate}
+ * @see {@link useWalletStatus}
+ */
+export function useIsWalletReady(): boolean {
+    const wallet = useWalletNamespace('useIsWalletReady');
+    const getIsReady = () => !isWalletWarmingUp(wallet.getState().status);
+    return useSyncExternalStore(wallet.subscribe, getIsReady, getIsReady);
 }
 
 // -- Action hooks -----------------------------------------------------------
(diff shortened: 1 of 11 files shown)


mcintyre94:
* **#319** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/react/index.ts
```diff
@@ -96,6 +98,34 @@ export function useWallets(): readonly UiWallet[] {
...
+ * ```tsx
+ * const isReady = useIsWalletReady();
+ * return <button disabled={!isReady}>Connect</button>;
+ * ```
+ *
+ * @see {@link WalletReadyGate}
+ * @see {@link useWalletStatus}
+ */
+export function useIsWalletReady(): boolean {
+    return !isWalletWarmingUp(useWalletStatus());
```
trevor-cortex: Minor perf: this composes `useWalletStatus()` (which selects the string via `useSyncExternalStore`) and derives a boolean, so `useIsWalletReady` / `WalletReadyGate` re-render on every status change even when `isReady` stays the same — e.g. `disconnected` → `connecting` → `connected` triggers three renders with `isReady === true` throughout. Not measurable in practice, but the other state hooks in this file are all careful about slice selection.

Could be tightened by giving the hook its own `useSyncExternalStore` with a boolean-returning `getSnapshot`:

```ts
export function useIsWalletReady(): boolean {
    const wallet = useWalletNamespace('useIsWalletReady');
    const getIsReady = () => !isWalletWarmingUp(wallet.getState().status);
    return useSyncExternalStore(wallet.subscribe, getIsReady, getIsReady);
}
```

Feel free to ignore if you'd rather keep the composition.

packages/kit-plugin-wallet/src/status.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @example
+ * ```ts
+ * client.wallet.subscribe(() => {
+ *     const { status } = client.wallet.getState();
+ *     setLoading(isWalletWarmingUp(status));
+ * });
+ * ```
+ *
+ * @see {@link whenReady} — the imperative (promise-based) counterpart.
```
trevor-cortex: Nit: `{@link whenReady}` isn't in scope here — it's a method on `WalletNamespace`. TypeDoc may still resolve it via the export graph, but `{@link WalletNamespace.whenReady}` (same for `useIsWalletReady` / `WalletReadyGate` if you want to symmetrise) would be unambiguous. Same nit applies to the `{@link whenReady}` reference in the `WalletReadyGate` docblock in `react/index.ts`.

trevor-cortex (commented):
Nice, tidy addition. The three-way split — `isWalletWarmingUp` predicate + imperative `whenReady()` + declarative `useIsWalletReady`/`WalletReadyGate` — all keyed off the same warm-up set is a clean design, and the docblocks are thorough about *why* user-initiated `connecting`/`disconnecting` are excluded.

### What this does

Adds first-class readiness helpers so apps can hide the initial auto-reconnect warm-up (`pending`/`reconnecting`) from wallet-dependent UI, without conflating it with user-initiated `connecting`/`disconnecting`. Exposes:
- `isWalletWarmingUp(status)` — the shared predicate.
- `client.wallet.whenReady()` — imperative, promise-based, Suspense-safe (stable reference while transient).
- `useIsWalletReady()` / `<WalletReadyGate fallback>` on the `/react` subpath.

Dispose semantics are handled: a store disposed mid-warm-up settles to `'disconnected'` and resolves any pending `whenReady()`, so a superseded client on a chain swap never leaves an `await` hanging.

### Things I liked

- Single source of truth (`isWalletWarmingUp`) shared by store + React + the docs' "framework integration" example. Easy to keep in sync later.
- `readyDeferred` is created lazily *inside* `whenReady()` only while transient, and cleared as soon as the status leaves the warm-up set. That means no allocation on the common already-settled path, and a fresh episode (if the state machine were ever extended to re-enter `pending`/`reconnecting`) would arm a fresh deferred cleanly.
- Disposal ordering in `Symbol.dispose`: `listeners.clear()` / `signerListeners.clear()` run *before* the final `updateState({ status: 'disconnected' })`, so subscribers don't get a spurious post-dispose notification. The inline comment calls this out, which is good — it's the kind of ordering that's easy to break in a future refactor.
- Test coverage is genuinely good: settled-immediately, stable-reference-across-a-warm-up-episode, resolves-on-settled-to-disconnected, resolves-on-dispose-mid-warm-up, and the post-dispose reconnect-can't-revive assertion. The `Promise.race([p.then(() => 'ready'), Promise.resolve('pending')])` trick to assert pending-without-timing-out is nice.
- SSR stub gets a `whenReady: () => Promise.resolve()` and a matching test. The stub's snapshot is permanently `'pending'`, so resolving immediately here is the right call (never hang on the server), even though it's technically inconsistent with `isWalletWarmingUp('pending')` — the SSR path is intentionally inert.

### Minor things worth considering (none blocking)

- **`useIsWalletReady` re-renders on every status change**, not only when the boolean flips. It composes `useWalletStatus()` (which selects the string) and derives a boolean, so a transition like `disconnected` → `connecting` → `connected` will re-render `WalletReadyGate` three times even though `isReady` stays `true` throughout. Straightforward to fix by giving the hook its own `useSyncExternalStore` with a boolean-returning `getSnapshot` (mirroring the other state hooks). Almost certainly not measurable in practice, but the other state hooks are all careful about slice selection, so it stands out a bit.
- **`@see {@link whenReady}` in `status.ts`** references the `WalletNamespace` method, not a symbol in scope. TypeDoc will probably still resolve it via the export graph, but `{@link WalletNamespace.whenReady}` would be unambiguous. Same nit for the `{@link whenReady}` in the `WalletReadyGate` docblock in `react/index.ts`.
- **README wording**: the intro to the `whenReady()` section says "the instant you build it" for how quickly the client is published — reads slightly awkward; "as soon as you build it" flows better. Purely stylistic.

### For subsequent reviewers

- The interesting invariant is in `updateState`: `readyDeferred` is only *ever* created while `isWalletWarmingUp(state.status)` is true (enforced by `whenReady()` itself), so the `isWalletWarmingUp(prev) && !isWalletWarmingUp(state)` edge is guaranteed to be the transition that owns any live deferred. Worth double-checking that no future code path calls `whenReady()`-adjacent logic from a non-transient state.
- The Suspense claim ("safe to throw for React Suspense") rests on `whenReady()` returning the *same* promise reference for the duration of one transient episode. The test explicitly asserts this (`expect(store.whenReady()).toBe(p)` both before and during `reconnecting`). If anyone later refactors `whenReady()` to allocate per-call, that guarantee breaks silently — worth guarding in a follow-up if this ever gets extracted into a shared plugin feature as the description hints.

(Overall: LGTM, would approve if I had write access.)

packages/kit-plugin-wallet/src/react/index.ts
mcintyre94: Nice, updated + added a test

packages/kit-plugin-wallet/src/status.ts
mcintyre94: Updated

lorisleiva (approved):
So clean! 🤤

mcintyre94:
### Merge activity

* **Jul 13, 2:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/319).
* **Jul 13, 2:33 PM UTC**: [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/319) rebased this pull request as part of a merge.
* **Jul 13, 2:33 PM UTC**: @mcintyre94 merged this pull request with [Graphite](https://app.graphite.com/github/pr/anza-xyz/kit-plugins/319).

PR #320, opened 2026-07-13
Fix publish: disable ESM splitting between root and react subpath

mcintyre94:
The version of the wallet plugin with the /react subpath does not currently work correctly, because it imports files from `chunk-` files that have not been published.

This is because `tsup` defaults ESM code splitting to true, which creates those chunks, and the `files` in our package.json does not include them.

This PR fixes the package by turning off this splitting. This works for the wallet plugin because the two paths mostly do not share code, only types. One caveat: the `isWalletWarmingUp` function is shared, and is now duplicated between both paths. Crucially they don't share any state, so there's no dual package hazard.

The alternative would be to extend `files` to include the chunks, but I think this is the cleaner option for this plugin.

For this plugin package only, a new test is added which checks for this issue on a dist/ directory.

Note that this only affects the latest version of the wallet plugin, because it's the only plugin with a subpath (in this case `/react`).

### DIFF
--- a/packages/kit-plugin-wallet/package.json
+++ b/packages/kit-plugin-wallet/package.json
@@ -64,7 +64,8 @@
     "scripts": {
         "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
         "dev": "vitest --project node",
-        "test": "pnpm test:types && pnpm test:treeshakability && pnpm test:unit",
+        "test": "pnpm test:types && pnpm test:treeshakability && pnpm test:packaging && pnpm test:unit",
+        "test:packaging": "node ../../scripts/assert-publishable-dist.mjs",
         "test:treeshakability": "for file in dist/index.*.mjs dist/react/index.*.mjs; do agadoo $file; done",
         "test:types": "tsc --noEmit",
         "test:unit": "vitest run"
--- a/scripts/assert-publishable-dist.mjs
+++ b/scripts/assert-publishable-dist.mjs
@@ -1,0 +1,103 @@
+#!/usr/bin/env node
+// Guards against publishing a package whose built entry points import files that
+// were not included in the npm tarball. This happens when a build emits shared
+// chunks (e.g. tsup code-splitting across multiple entries) but the package's
+// `files` allow-list only matches the entry files, so the chunks silently drop
+// out of the published tarball and every ESM consumer fails to resolve them.
+// @see https://github.com/anza-xyz/kit-plugins — the `@solana/kit-plugin-wallet@0.13.0` incident.
+//
+// Usage: node scripts/assert-publishable-dist.mjs [packageDir]
+// Runs against the current working directory by default.
+
+import { execFileSync } from 'node:child_process';
+import { existsSync, readFileSync } from 'node:fs';
+import { dirname, join, posix, relative, resolve } from 'node:path';
+import { argv, cwd, exit } from 'node:process';
+
+const packageDir = resolve(argv[2] ?? cwd());
+
+/** Ask npm which files this package would actually publish. */
+function getPublishedFiles() {
+    const output = execFileSync('npm', ['pack', '--dry-run', '--json'], {
+        cwd: packageDir,
+        encoding: 'utf8',
+        stdio: ['ignore', 'pipe', 'ignore'],
+    });
+    // `npm pack --json` paths are package-root-relative, POSIX-separated, no `package/` prefix.
+    return new Set(JSON.parse(output)[0].files.map(f => f.path));
+}
+
+/** Extract every relative (`./` or `../`) module specifier referenced by a JS module. */
+function getRelativeSpecifiers(code) {
+    const specifiers = new Set();
+    // Matches: import/export ... from '…', bare `import '…'`, dynamic import('…'), and require('…').
+    const patterns = [
+        /(?:\bimport\b|\bexport\b)[^;'"]*?\bfrom\s*['"]([^'"]+)['"]/g,
+        /\bimport\s*['"]([^'"]+)['"]/g,
+        /\bimport\s*\(\s*['"]([^'"]+)['"]\s*\)/g,
+        /\brequire\s*\(\s*['"]([^'"]+)['"]\s*\)/g,
+    ];
+    for (const pattern of patterns) {
+        for (const [, specifier] of code.matchAll(pattern)) {
+            if (specifier.startsWith('.')) {
+                specifiers.add(specifier);
+            }
+        }
+    }
+    return specifiers;
+}
+
+/** Resolve a relative specifier (from `fromFile`) to a package-root-relative POSIX path, or null. */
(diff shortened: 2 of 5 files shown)


mcintyre94:
@trevor-cortex

scripts/assert-publishable-dist.mjs
```diff
@@ -0,0 +1,103 @@
...
+
+/** Extract every relative (`./` or `../`) module specifier referenced by a JS module. */
+function getRelativeSpecifiers(code) {
+    const specifiers = new Set();
+    // Matches: import/export ... from '…', bare `import '…'`, dynamic import('…'), and require('…').
+    const patterns = [
+        /(?:\bimport\b|\bexport\b)[^;'"]*?\bfrom\s*['"]([^'"]+)['"]/g,
+        /\bimport\s*['"]([^'"]+)['"]/g,
+        /\bimport\s*\(\s*['"]([^'"]+)['"]\s*\)/g,
+        /\brequire\s*\(\s*['"]([^'"]+)['"]\s*\)/g,
```
trevor-cortex, line 38: Minor: regex-based specifier extraction over raw source is a heuristic — it can both miss real imports (unusual whitespace/formatting) and produce false positives (e.g. an `import '...'` inside a comment or template literal). For tsup-generated ESM/CJS output this is almost certainly fine in practice, but worth calling out as a known limitation in case someone later extends this to check hand-written code or other bundlers' output. A parse-based approach (`es-module-lexer` / `acorn`) would be more robust if this ever becomes a problem.

scripts/assert-publishable-dist.mjs
```diff
@@ -0,0 +1,103 @@
...
+
+import { execFileSync } from 'node:child_process';
+import { existsSync, readFileSync } from 'node:fs';
+import { dirname, join, posix, relative, resolve } from 'node:path';
+import { argv, cwd, exit } from 'node:process';
+
+const packageDir = resolve(argv[2] ?? cwd());
+
+/** Ask npm which files this package would actually publish. */
+function getPublishedFiles() {
```
trevor-cortex, line 20: nit: `stdio: ['ignore', 'pipe', 'ignore']` swallows npm's stderr. If `npm pack --dry-run` ever emits a genuinely useful warning (deprecated engines, missing fields, etc.) it'll be silently discarded, and if the command itself fails the error surface will be pretty terse. Consider `'inherit'` for stderr, or capturing it and re-emitting on non-zero exit.

packages/kit-plugin-wallet/package.json
```diff
@@ -64,7 +64,8 @@
     "scripts": {
         "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
         "dev": "vitest --project node",
-        "test": "pnpm test:types && pnpm test:treeshakability && pnpm test:unit",
+        "test": "pnpm test:types && pnpm test:treeshakability && pnpm test:packaging && pnpm test:unit",
+        "test:packaging": "node ../../scripts/assert-publishable-dist.mjs",
```
trevor-cortex, line 68: The `test:packaging` script is generic — nothing about `assert-publishable-dist.mjs` is wallet-specific. Consider adding it to the default package test recipe (or a shared preset) so future packages that grow subpath exports get the check automatically, rather than having to remember to opt in. Not a blocker for this PR; happy to leave it as follow-up.

tsup.config.base.ts
```diff
@@ -38,6 +38,12 @@ export function getBuildConfig(options: BuildOptions): TsupConfig {
         publicDir: true,
         pure: ['process'],
         sourcemap: true,
+        // Keep every entry point self-contained. tsup defaults ESM `splitting` to `true`, which
+        // hoists code shared between entries (e.g. `./index` and `./react`) into `chunk-*` files.
+        // Those chunks are not matched by packages' `files` allow-lists, so they silently drop out
+        // of the published tarball.
+        // @see scripts/assert-publishable-dist.mjs
+        splitting: false,
```
trevor-cortex, line 46: The comment here is great — clearly explains why `splitting: false` is intentional and points readers to the guardrail. One tiny suggestion: worth noting that this is a global setting affecting every package, and that the wallet plugin is the only one where it currently changes anything (single-entry packages get `splitting: true` for free). Helps future readers understand the blast radius when they're considering re-enabling it.

trevor-cortex (commented):
Clean, well-motivated fix for the broken `@solana/kit-plugin-wallet@0.13.0` publish. LGTM overall.

## What this does

- Sets `splitting: false` in the shared tsup base config so every entry point is self-contained. This is the actual bug fix: tsup's ESM default (`splitting: true`) was hoisting code shared between `./` and `./react` into `chunk-*.mjs` files, which the wallet package's `files` allow-list (`dist/index.*`, `dist/react/index.*`) never matched — so the chunks silently dropped out of the published tarball and every ESM consumer's `import './chunk-*.mjs'` failed to resolve.
- Adds `scripts/assert-publishable-dist.mjs`, a package-agnostic guardrail that runs `npm pack --dry-run --json`, walks every published JS entry point, and asserts that every relative import resolves to a file that is also in the tarball.
- Wires the guardrail up as `test:packaging` on `@solana/kit-plugin-wallet` and adds a matching `turbo.json` task (correctly depending on `build`, not `^build`).
- Ships a `patch` changeset — appropriate for a publish fix.

## Things to watch out for

- **`splitting: false` is now global** — it applies to every package via the shared base, not just the wallet plugin. This is fine in practice because every other plugin has a single entry point, so splitting was a no-op for them anyway. The only observable effect is on the wallet plugin, where `isWalletWarmingUp` (a pure predicate, no shared state — verified in `src/status.ts`) is now duplicated between the two entry points. No dual-package hazard risk, and the size impact is negligible.
- **The guardrail is only wired into the wallet package.** Defensible today (it's the only package with multiple entries, and therefore the only one where splitting could ever produce chunks), but the script is generic — worth considering whether to fold it into the default `test` recipe so future packages that add subpath exports get the check for free. Not a blocker.
- **The changeset copy is very detailed** — great context for the changelog. Verified the reasoning against `packages/kit-plugin-wallet/package.json`'s `files` field and `tsup.config.ts` entries; the narrative matches reality.

- Worth double-checking that `test:packaging` actually reproduces the bug against the pre-fix state (i.e. temporarily flip `splitting` back to `true`, rebuild, and confirm the script fails with the expected `NOT in the published tarball` message). The regex-based specifier extraction is a heuristic, so a smoke test that it catches the exact class of bug it's designed to catch is valuable.
- The script's specifier extraction uses regex over source text rather than a real parser (see inline comment). For tsup-generated output that's almost certainly fine, but flagging it as a known limitation.
- No changes to the wallet plugin's public API or docs — this is a pure packaging/build fix, so no README or docblock updates are needed.

(Leaving as `COMMENT` rather than `APPROVE` because I don't have write access on this repo.)

lorisleiva (approved):
Thank you for fixing this.

PR #326, opened 2026-07-14
Update wallet hooks to require the client as the first param

mcintyre94:
This PR updates all the react hooks in the wallet plugin to take `client: ClientWithWallet` as their first param. This removes the runtime type checking with `useClientCapability`, and moves the responsibility to the app. This allows the app to have proper type checking all the way through plugin hooks. by typing its `useClient` parameter.

This is a breaking change, but this is a more robust and type-safe pattern and it's worth getting it right now instead of repeating the `useClientCapability` pattern through other hooks.

This PR updates the docs in this repo, I'll follow up with a Kit PR that updates the docs there to talk about passing a client type around. We don't currently have these hooks documented in Kit, I've opened an issue to do that separately.

```diff
- const status = useWalletStatus();
- const { dispatch } = useConnect();
+ const status = useWalletStatus(client);
+ const { dispatch } = useConnect(client);
```

```diff
- <WalletReadyGate fallback={<Spinner />}>
+ <WalletReadyGate client={client} fallback={<Spinner />}>
      <WalletDependentUI />
  </WalletReadyGate>
```

Additional context: https://github.com/anza-xyz/kit/pull/1833#discussion_r3572581679 - hooks that use `useClientCapability` limit type safety and undo some of the benefits of plugins. Having the app pass `client` allows the app to maintain type safety.

### DIFF
--- a/.changeset/social-peaches-doubt.md
+++ b/.changeset/social-peaches-doubt.md
@@ -1,0 +1,19 @@
+---
+'@solana/kit-plugin-wallet': minor
+---
+
+The React hooks now take the wallet-enabled `client` as their first argument instead of reading it from a `ClientProvider` context. Every hook — `useWalletStatus`, `useConnectedWallet`, `useWallets`, `useIsWalletReady`, `useConnect`, `useDisconnect`, `useSignIn`, `useSignMessage`, `useSelectAccount` — and the `WalletReadyGate` component (via a new `client` prop) now accept `client: ClientWithWallet` directly. Because the client's type already guarantees the wallet plugin is installed, the runtime capability assertion and the provider requirement are gone. This is a breaking change: pass the client into every hook and into `WalletReadyGate`:
+
+```diff
+- const status = useWalletStatus();
+- const { dispatch } = useConnect();
++ const status = useWalletStatus(client);
++ const { dispatch } = useConnect(client);
+```
+
+```diff
+- <WalletReadyGate fallback={<Spinner />}>
++ <WalletReadyGate client={client} fallback={<Spinner />}>
+      <WalletDependentUI />
+  </WalletReadyGate>
+```
--- a/packages/kit-plugin-wallet/README.md
+++ b/packages/kit-plugin-wallet/README.md
@@ -171,7 +171,7 @@ All wallet state is accessed via `client.wallet.getState()`, which returns a ref
 
 ## React hooks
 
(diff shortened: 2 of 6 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Clean, well-scoped refactor — LGTM. Every React hook and `WalletReadyGate` now take `client: ClientWithWallet` as their first argument, replacing the runtime `useClientCapability` check with a static TypeScript constraint. Because `ClientWithWallet` already encodes that a wallet plugin is installed, both the capability assertion and the `ClientProvider` requirement fall away, and app code gets end-to-end type-checking through the plugin.

### What to look at

- **Public API surface** (`src/react/index.ts`): every hook signature changes, plus `WalletReadyGateProps` gains a required `client` prop — all breaking, correctly captured in the changeset.
- **Docblocks**: `@param client` added consistently; the old `Requires a ClientProvider… asserts this at mount with useClientCapability` boilerplate is removed everywhere. Skimmed all nine — no stragglers.
- **Type tests** (`__typetests__/index.ts`): now enforce that `client` is required on every hook and on `WalletReadyGateProps` via `@ts-expect-error`. These replace the deleted `throws at mount when the client lacks a wallet namespace` runtime tests, which is the right tradeoff — the check moved from runtime to compile time, so the coverage moves with it.
- **README**: intro paragraph, `WalletButton` example, and the warm-up gating example are all updated to thread `client: ClientWithWallet` through. Grep is clean — no lingering `ClientProvider` references in the React hooks section.
- **Changeset**: `minor` bump is correct for pre-1.0 (`0.13.1`) per this repo's changesets skill (pre-1.0 minor bumps may be treated as breaking).

### Notes for subsequent reviewers

- `useSelectAccount` now returns `client.wallet.selectAccount` directly (no `useSyncExternalStore`, no closure). Since `selectAccount` is a stable method on the namespace, the existing "stable across renders" test still holds — no hidden regression.
- State hooks pass `wallet.subscribe` to `useSyncExternalStore`. That reference is stable on the wallet namespace, so callers won't see extra resubscribes as long as they pass the same `client`.
- Worth double-checking the follow-up Kit docs PR mentioned in the description lands in lockstep, since the two READMEs currently disagree on the hook signatures.

lorisleiva (approved):
Thanks! A little more verbose but it pushes the Client type at the forefront of the API rather than hiding it until something fails at runtime. 👌

PR #341, opened 2026-07-23
[kit-plugin-litesvm] Add `GetProgramAccountsApi` support to litesvm Rpc

elliotkennedy:
litesvm now exposes a `get_program_accounts` API (https://github.com/LiteSVM/litesvm/pull/370), which we can use to implement `GetProgramAccountsApi` on the LiteSVM-backed Rpc.                                                                       
                                                                                                                                                                                                                                                    
The implementation supports `dataSize`/`memcmp` filters, `dataSlice`, and `withContext`. Account data is returned as base64 only - consistent with the existing `getAccountInfo`/`getMultipleAccounts` methods, which reject other encodings.

### DIFF
--- a/packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
+++ b/packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
@@ -3,20 +3,29 @@ import {
     AccountInfoWithBase64EncodedData,
     Address,
     Base64EncodedBytes,
+    containsBytes,
     createJsonRpcApi,
     createRpc,
+    DataSlice,
+    EncodedAccount,
     GetAccountInfoApi,
     GetBalanceApi,
     getBase58Decoder,
+    getBase58Encoder,
     getBase64Decoder,
+    getBase64Encoder,
     GetEpochScheduleApi,
     GetLatestBlockhashApi,
     GetMinimumBalanceForRentExemptionApi,
     GetMultipleAccountsApi,
+    GetProgramAccountsApi,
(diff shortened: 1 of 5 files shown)


packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
```diff
@@ -166,3 +193,44 @@ function assertEncodingIsBase64(encoding: Encoding | undefined) {
...
+        const start = Number(offset);
+        if (start + needle.length > data.length) {
+            return false;
+        }
+        for (let i = 0; i < needle.length; i++) {
+            if (data[start + i] !== needle[i]) {
+                return false;
+            }
+        }
+        return true;
```
lorisleiva: I believe this could be simplified by using the [`containsByte`](https://www.solanakit.com/api/functions/containsBytes) function from `@solana/kit`.

lorisleiva (approved):
Love it! Thanks for adding this to LiteSVM and Kit Plugins. I've just got a small comment on reusing a Kit primitive to avoid repetition. 🙂

packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
elliotkennedy: Good shout - refactored

elliotkennedy:
Hey @lorisleiva - thanks for the review + this awesome lib - please take another look when you can

lorisleiva:
Thank you! 🍺

PR #344, opened 2026-07-27
Add configurable compute unit buffer and resource limit estimation controls

lorisleiva:
This PR reworks resource limit estimation for the RPC plugin and prepares the transaction planner config for version 1 transactions.

The compute unit buffer applied to simulation estimates is now configurable through a new `getComputeUnitLimitFromEstimate` option on `rpcTransactionPlanExecutor` (surfaced on `solanaRpc`). The previous flat 10% buffer is replaced by a default that enforces a 300 CU floor and a margin decaying linearly from 10% to 2% by 500,000 CU, with the result always capped at the 1,400,000 per-transaction maximum. Resource limit estimation can now be disabled with `estimateResourceLimits: false` for size-sensitive transactions.

`TransactionPlannerConfig` on both the RPC and LiteSVM plugins becomes a version-discriminated union so that version 1 support (structured resource headers) can land later without a breaking change; passing `version: 1` throws for now since `@solana/kit` cannot yet build v1 messages.

### DIFF
--- a/packages/kit-plugin-litesvm/src/transaction-planner.ts
+++ b/packages/kit-plugin-litesvm/src/transaction-planner.ts
@@ -3,6 +3,7 @@ import {
     createTransactionMessage,
     createTransactionPlanner,
     extendClient,
+    Lamports,
     MicroLamports,
     pipe,
     setTransactionMessageComputeUnitPrice,
@@ -15,6 +16,10 @@ import {
  * The planner creates transaction messages with:
  * - The configured fee payer.
  * - Optional priority fees.
+ *
+ * Unlike the RPC planner, the LiteSVM planner does not estimate resource limits,
+ * since LiteSVM executes transactions locally without a simulation-based
+ * estimation step.
  *
  * @param config - Optional configuration for the planner.
  * @returns A plugin that adds `transactionPlanner` to the client.
@@ -34,15 +39,28 @@ import {
  */
 export function litesvmTransactionPlanner(config: TransactionPlannerConfig = {}) {
     return <T extends ClientWithPayer>(client: T) => {
+        if (config.version === 1) {
+            // The v1 transaction path is defined at the type level for forward
+            // compatibility, but `createTransactionMessage` cannot yet build v1
+            // messages, so we fail loudly rather than silently misbehave.
+            throw new Error(
+                'Version 1 transactions are not yet supported by `litesvmTransactionPlanner`. ' +
+                    'Use version 0 or legacy transactions for now.',
+            );
+        }
+
+        const version = config.version ?? 0;
+        const microLamportsPerComputeUnit = config.microLamportsPerComputeUnit;
+
         const transactionPlanner = createTransactionPlanner({
             createTransactionMessage: () => {
                 return pipe(
-                    createTransactionMessage({ version: config.version ?? 0 }),
+                    createTransactionMessage({ version }),
                     tx => setTransactionMessageFeePayerSigner(client.payer, tx),
                     tx =>
-                        config.microLamportsPerComputeUnit
-                            ? setTransactionMessageComputeUnitPrice(config.microLamportsPerComputeUnit, tx)
-                            : tx,
+                        microLamportsPerComputeUnit === undefined
+                            ? tx
+                            : setTransactionMessageComputeUnitPrice(microLamportsPerComputeUnit, tx),
                 );
             },
         });
@@ -52,18 +70,17 @@ export function litesvmTransactionPlanner(config: TransactionPlannerConfig = {})
 }
 
 /**
- * Configuration options for the transaction planner.
+ * Configuration options for the transaction planner when creating legacy or
+ * version 0 transaction messages.
  *
- * The `version` field is used to determine the transaction version
- * to use when creating transaction messages and determines the shape
- * of the rest of the configuration options, as some options are only
- * applicable to certain transaction versions.
+ * For these versions, priority fees are expressed as a `setComputeUnitPrice`
+ * compute budget instruction appended to the transaction.
  */
-// TODO(loris): Add support for v1 transactions when LiteSVM supports them.
-// This includes: `version: 1` and `priorityFees?: Lamports` in a new union variant.
-export type TransactionPlannerConfig = {
+export type TransactionPlannerConfigLegacy = {
     /**
-     * The priority fees to be set on the transaction in micro lamports per compute unit.
+     * The priority fee to set on the transaction, in micro-lamports per compute
+     * unit, added as a `setComputeUnitPrice` instruction.
+     *
      * Defaults to using no priority fees.
      */
     microLamportsPerComputeUnit?: MicroLamports;
@@ -73,3 +90,42 @@ export type TransactionPlannerConfig = {
      */
     version?: 'legacy' | 0;
 };
+
+/**
+ * Configuration options for the transaction planner when creating version 1
+ * transaction messages.
+ *
+ * For version 1, priority fees live in a structured resource header rather than
+ * a compute budget instruction, and are expressed as a total amount in lamports
+ * rather than a per-compute-unit price.
+ *
+ * @remarks
+ * Version 1 transactions are not yet buildable by `@solana/kit`, so this branch
+ * is currently inert: passing `version: 1` throws at runtime. The shape is
+ * defined now so that enabling version 1 later is not a breaking change.
+ */
+export type TransactionPlannerConfigV1 = {
+    /**
+     * The total priority fee to set on the transaction, in lamports, written to
+     * the version 1 resource header.
+     *
+     * Defaults to using no priority fees.
+     */
+    priorityFeeLamports?: Lamports;
+    /**
+     * The transaction message version to use when creating transaction messages.
+     */
+    version: 1;
+};
+
+/**
+ * Configuration options for the transaction planner.
+ *
+ * The `version` field determines the transaction version to use when creating
+ * transaction messages and discriminates the shape of the rest of the
+ * configuration, as some options are only applicable to certain versions.
+ *
+ * @see {@link TransactionPlannerConfigLegacy}
+ * @see {@link TransactionPlannerConfigV1}
+ */
+export type TransactionPlannerConfig = TransactionPlannerConfigLegacy | TransactionPlannerConfigV1;
--- a/packages/kit-plugin-rpc/src/transaction-plan-executor.ts
+++ b/packages/kit-plugin-rpc/src/transaction-plan-executor.ts
@@ -50,6 +50,43 @@ import {
 export function rpcTransactionPlanExecutor(
     config: {
         /**
+         * Whether to estimate and set resource limits (the compute unit limit
+         * and, for version 1 transactions, the loaded accounts data size limit)
+         * by simulating the transaction before sending.
+         *
+         * When `true` (default), any resource limit that is still provisory or
+         * unset is estimated via a simulation and replaced with the estimated
+         * value (with a compute unit buffer applied via
+         * `getComputeUnitLimitFromEstimate`). When `false`, no estimation
+         * simulation is performed and the message is sent with exactly the
+         * resource limits it already carries.
+         *
+         * This should match the `estimateResourceLimits` option passed to
+         * {@link rpcTransactionPlanner}, which controls whether provisory limits
+         * are reserved in the first place. {@link solanaRpc} keeps them in sync
+         * automatically.
+         *
+         * Defaults to `true`.
+         */
+        estimateResourceLimits?: boolean;
+        /**
+         * Maps the estimated compute unit consumption from simulation to the
+         * compute unit limit to set on the transaction, adding headroom to
+         * account for variation between simulation and execution.
+         *
+         * By default, a function is used that adds a buffer on top of the
+         * estimate of at least 300 compute units, or a margin that decays
+         * linearly from 10% at low estimates to 2% at 500,000 compute units and
+         * above, whichever is greater.
+         *
+         * The returned value is always capped at 1,400,000, the maximum number of
+         * compute units allowed per transaction.
+         *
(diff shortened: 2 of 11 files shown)


lorisleiva:
* **#344** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/344?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/344?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR does three related things on the RPC transaction plan executor / planner (and the LiteSVM planner, for consistency):

1. **Configurable compute unit buffer.** The previous flat 10% buffer becomes a `getComputeUnitLimitFromEstimate: (estimatedComputeUnits: number) => number` option on `rpcTransactionPlanExecutor` (and `solanaRpc`). The new default enforces a 300 CU floor and a margin decaying linearly from 10% at low estimates to 2% at 500,000+ CU, with the final result always capped at 1,400,000. The buffer is applied on both the success path and the `skipPreflight` recovery path.
2. **`estimateResourceLimits: false` opt-out.** Threaded through the planner (skips `fillTransactionMessageProvisoryResourceLimits`), the executor (skips the estimation simulation entirely), and `solanaRpc` (single source in `transactionConfig`, wired to both).
3. **`TransactionPlannerConfig` as a version-discriminated union.** Adds `TransactionPlannerConfigLegacy` and `TransactionPlannerConfigV1` on both plugins, so that a future `priorityFeeLamports` (v1 resource header) can land without a breaking change. `version: 1` throws at runtime for now.

The math on the default buffer curve checks out (spot-checked 50,000 → 54,600, 42 → 300 via the floor, and 1,390,000 → 27,800 extra → 1,417,800 → capped at 1,400,000). Tests cover the small/mid/cap/custom-fn/recovery-path cases and the type-level discrimination on both plugins. Docs on `solanaRpc`, the planner, and the executor are kept in sync, and the readme sections have been updated.

## For subsequent reviewers

- **Planner/executor sync when not using `solanaRpc`.** `solanaRpc` mirrors `estimateResourceLimits` to both plugins automatically, but a user wiring `rpcTransactionPlanner` and `rpcTransactionPlanExecutor` by hand can easily set one and not the other. If the planner reserves provisory limits (default) but the executor is told `estimateResourceLimits: false`, the transaction goes out with a provisory CU limit of 0. The executor's docblock calls this out, but it's an easy footgun — worth confirming this is acceptable rather than something we want to guard against.
- **Umbrella re-export asymmetry.** `@solana/kit-plugins` now re-exports `TransactionPlannerConfigLegacy` / `TransactionPlannerConfigV1` from `@solana/kit-plugin-rpc` only. The LiteSVM variants have the same names but a strictly smaller shape (no `estimateResourceLimits`). The comment calls them supersets, which is accurate, but anyone importing these types from the umbrella and passing them to `litesvmTransactionPlanner` will be able to set `estimateResourceLimits` at the type level even though it's ignored at runtime. Probably fine given the umbrella is deprecated, but worth being aware of.
- **V1 type shape is speculative.** `priorityFeeLamports: Lamports` on `TransactionPlannerConfigV1` locks in a name and unit for a feature that isn't buildable yet. If Kit lands v1 support with a different field name or additional required fields, this becomes a soft breaking change even though the branch currently throws. The `@remarks` docblock is upfront about this — flagging so it's a deliberate choice.
- **Recovery path with missing `unitsConsumed`.** On the `skipPreflight` recovery path, `error.context.unitsConsumed ?? 0n` means a failed simulation with no reported units resolves to `estimatedComputeUnits = 0`, which the default function turns into the 300 floor. That seems like reasonable behaviour (transaction still reaches the validator with a tiny CU limit for debugging), but worth verifying it matches the intent.

Overall this looks good — clean split of concerns, tests are thorough, and the docs are careful about the interactions between `estimateResourceLimits`, `skipPreflight`, and preflight. Just leaving comments; no blocking issues.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -138,30 +187,76 @@ export function rpcTransactionPlanExecutor(
...
+ * @param estimatedComputeUnits - The compute units consumed during simulation.
+ * @returns The compute unit limit to set on the transaction.
+ */
+function getDefaultComputeUnitLimitFromEstimate(estimatedComputeUnits: number): number {
+    const progress = Math.min(estimatedComputeUnits / COMPUTE_UNIT_MARGIN_CAP, 1);
+    const margin = MAX_COMPUTE_UNIT_MARGIN - (MAX_COMPUTE_UNIT_MARGIN - MIN_COMPUTE_UNIT_MARGIN) * progress;
+    // Compute the extra units separately and round them up before adding, so we
+    // avoid floating-point artefacts from multiplying by `1 + margin` directly.
+    const extraComputeUnits = Math.ceil(estimatedComputeUnits * margin);
+    return estimatedComputeUnits + Math.max(MIN_COMPUTE_UNIT_BUFFER, extraComputeUnits);
```
amilz, line 228: I think we want to run `Math.ceil` on the full result to make sure it's an integer.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -196,10 +291,14 @@ function bufferAndRecoverResourceLimits(
...
         onSimulate();
 
-        // Multiply the estimated compute unit limit by 1.1 to add a 10% buffer.
+        // Apply the compute unit buffer to the estimated compute unit limit,
+        // capping at the per-transaction maximum.
         return {
             ...estimate,
-            computeUnitLimit: Math.ceil(estimate.computeUnitLimit * 1.1),
+            computeUnitLimit: Math.min(
+                MAX_COMPUTE_UNIT_LIMIT,
```
amilz, line 300: I understand why `MAX_COMPUTE_UNIT_LIMIT` is here as a protection rather than inside of `getComputeUnitLimitFromEstimate` -- however, it might make sense to move it into `getComputeUnitLimitFromEstimate` so if someone is testing a feature gate or cluster that has a higher ceiling, they'd be able to include that in their custom config.

packages/kit-plugin-rpc/src/transaction-planner.ts
```diff
@@ -34,19 +36,41 @@ import {
...
-                    createTransactionMessage({ version: config.version ?? 0 }),
+                    createTransactionMessage({ version }),
                     tx => setTransactionMessageFeePayerSigner(client.payer, tx),
-                    tx => fillTransactionMessageProvisoryResourceLimits(tx),
+                    tx => (estimateResourceLimits ? fillTransactionMessageProvisoryResourceLimits(tx) : tx),
                     tx =>
-                        config.microLamportsPerComputeUnit
-                            ? setTransactionMessageComputeUnitPrice(config.microLamportsPerComputeUnit, tx)
-                            : tx,
+                        microLamportsPerComputeUnit === undefined
```
amilz, line 71: Did you want to add a default value?

packages/kit-plugin-rpc/test/transaction-plan-executor.test.ts
amilz, line 1: wdyt about a `it('performs no simulation when estimation is disabled and preflight is skipped', async () => {` and `expect(simulateTransaction).not.toHaveBeenCalled();`

amilz (commented):
LGTM. Couple of nits and a question about default pri fee.

packages/kit-plugin-rpc/src/transaction-planner.ts
lorisleiva: Hmm maybe I got this wrong but if I remember correctly we decided together not to add any default value for priority fees anymore since the new CU function accounts for the extra instruction now. The benefit of this is we don't have to guess a good enough `microLamportsPerComputeUnit` and more importantly we don't need to guess a good enough "flat" `priorityFee` amount when transactions v1 are out. It also makes the API more intuitive because now you don't need to opt-out of priority fees. What do you reckon?

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
lorisleiva: I'd still lean on keeping the global `MAX_COMPUTE_UNIT_LIMIT` protection for the following reasons:
- These plugins are designed for the 90% use-cases which will create transactions within the currently available limits. With that in mind, I think there's more value in protecting a user that offer a function such as `cu => cu * 1.5` versus allowing some advance user to bypass the current limits to test some upcoming feature.
- Advanced users testing new limits can easily bypass this by providing an explicit CU value themselves on the transaction.

packages/kit-plugin-rpc/test/transaction-plan-executor.test.ts
lorisleiva: Good shout! I've added the test.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
lorisleiva: Good call, the default function is actually already safe (`extraComputeUnits` is already using `Math.ceil`) but the call site that also accepts a custom function isn't protected. I've added a `Math.ceil` there.

```diff
- getComputeUnitLimitFromEstimate(estimate.computeUnitLimit),
+ Math.ceil(getComputeUnitLimitFromEstimate(estimate.computeUnitLimit))
```

packages/kit-plugin-litesvm/src/transaction-planner.ts
```diff
@@ -34,15 +39,28 @@ import {
...
 export function litesvmTransactionPlanner(config: TransactionPlannerConfig = {}) {
     return <T extends ClientWithPayer>(client: T) => {
+        if (config.version === 1) {
+            // The v1 transaction path is defined at the type level for forward
+            // compatibility, but `createTransactionMessage` cannot yet build v1
+            // messages, so we fail loudly rather than silently misbehave.
+            throw new Error(
+                'Version 1 transactions are not yet supported by `litesvmTransactionPlanner`. ' +
+                    'Use version 0 or legacy transactions for now.',
+            );
```
mcintyre94, line 49: I think this is probably the right call, but FWIW Kit can fully build and send v1 messages with the current version. The only thing missing is the `createTransactionMessage` version, ie you'd get a type error on that line if you didn't have this narrowing. And obviously network support, which is why the type change hasn't been made to Kit yet. So debatable, but if we decide to add it to the type here now (forward compatibility is a fair argument for that), we could also just allow it. You'd just need to `ts-ignore` the `createTransactionMessage` line.

packages/kit-plugin-rpc/src/transaction-planner.ts
mcintyre94: I'd agree with this - I don't think there's a good way to set a default (on mainnet), and the same with the flat `priorityFee` for v1. Something we could think about in future is that the new RPC spec should eventually include a fee estimate API (if I understand correctly), which we could add to Kit and potentially use in a similar way to simulation to estimate CUs to automatically set a default.

packages/kit-plugins/src/index.ts
```diff
@@ -3,9 +3,14 @@ export * from '@solana/kit-plugin-instruction-plan';
...
-export type { TransactionPlannerConfig } from '@solana/kit-plugin-rpc';
+// Resolve ambiguous re-exports. Both `@solana/kit-plugin-litesvm` and
+// `@solana/kit-plugin-rpc` export these types with the same name. We re-export
+// the `@solana/kit-plugin-rpc` variants, which are supersets of the LiteSVM
+// ones (they additionally support resource limit estimation).
+export type {
+    TransactionPlannerConfig,
+    TransactionPlannerConfigLegacy,
+    TransactionPlannerConfigV1,
+} from '@solana/kit-plugin-rpc';
```
mcintyre94, line 14: Nit: I'd lean toward not updating this deprecated package

mcintyre94 (approved):
LGTM! The new default function is much better and being able to override it is much nicer.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
mcintyre94: I wonder if it'd be worth just adding a `console.warn` if the limit is over `MAX_COMPUTE_UNIT_LIMIT` just so you know to bypass the plugin here. Also might be useful for people writing these CU estimate functions, maybe.

packages/kit-plugin-litesvm/src/transaction-planner.ts
lorisleiva: Good to know! Do we also get a useful runtime error though or would we just get an error if we tried to send it (if I refactored this to just ignore the type error)?

packages/kit-plugins/src/index.ts
lorisleiva: Yeah, Claude keeps rightfully updating these deprecated packages but I just need to pull the plug and remove them now. They've been deprecated for long enough now.

packages/kit-plugin-litesvm/src/transaction-planner.ts
mcintyre94: There's no runtime check for the version in Kit, so you'd just get the RPC error, ie no probably not useful! Runtime error here and then remove it when we decide to add the type to Kit (should be alongside the Agave 4.2 release) probably makes the most sense I think.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
lorisleiva: As discussed offline, let's not do a `console.warn` since that would be the first time we do this in both Kit and Kit Plugins. We also noted that Kit itself does the same clamping in its helper functions without any warnings.

PR #353, opened 2026-08-03
Fix broken deprecation notice links in READMEs

lorisleiva:
This PR fixes the cross-package links in the five deprecated packages' README deprecation notices. They previously used relative paths (e.g. `../kit-plugin-rpc`), which npm rewrites into invalid git-ref URLs like `github.com/anza-xyz/kit-plugins/blob/kit-plugin-rpc` and render as 404s on the package pages. Each link now uses an absolute URL pinned to commit `497b55f`, so it resolves correctly on npm and won't break once these deprecated packages are removed from `main`. It also adds the `[DEPRECATED]` prefix to the `kit-client-rpc` and `kit-client-litesvm` package descriptions for consistency with the other deprecated packages, and includes a patch changeset to republish the corrected READMEs.


lorisleiva:
* **#353** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/353?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/353?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #355, opened 2026-08-03
Add a solanaTestnetRpc plugin

mcintyre94:
This is modeled from solanaDevnetRpc, adds the full RPC stack typed for testnet, and includes airdrop


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Clean addition that mirrors `solanaDevnetRpc` faithfully — same signature, same defaulting pattern, same `rpcAirdrop()` bundling, just typed as `TestnetUrl` and pointed at `https://api.testnet.solana.com`. Docblocks, README section, root README plugin table, and cross-referenced `@see` tags on all sibling functions are all in place, and the changeset is a `minor` bump on `@solana/kit-plugin-rpc` which fits a new public API. LGTM.

### Things to watch out for

- Nothing structural. `TestnetUrl` is already exported from `@solana/kit` and the cast pattern matches devnet exactly.

### Notes for subsequent reviewers

- The `accepts custom config overrides` test only asserts that `rpc` is present, not that the overridden `rpcUrl` was actually forwarded. That's the same shape as the existing devnet test, so it's consistent — worth noting if you want to strengthen both later, but not something to block this PR on.

lorisleiva (approved):

PR #361, opened 2026-08-10
Refactor plan and send transaction plugin functions

mcintyre94:
We plan to add functions for `client.signTransaction(s)`, which will be backed by an independent transaction plan executor. This PR addresses 2 concerns with this: first, deprecating the single `client.transactionPlanExecutor`, and secondly re-working how we add the plan and send functions to be more extensible.

- The `transactionPlanner` plugin now adds `planTransaction(s)` functions
- The `transactionPlanExecutor` plugin is deprecated
- It's replaced by a new `transactionPlanSendingExecutor`, which adds `sendTransaction(s)` functions, and consumes the `planTransaction(s)` functions already on the client
- `transactionPlanner` and `transactionPlanExecutor` are still written to the client, but both are deprecated. The `plan/sendTransaction(s)` functions should be used instead.
- In the `rpc` and `litesvm` packages, we refactor plugins like `rpcTransactionPlanner` to call the `transactionPlanner` plugin from the instruction-plans package. This is necessary because the instruction-plans plugin now adds the actual `planTransaction(s)` functions. Previously it just wrote `transactionPlanner` to the client and `kit-plugin-rpc` just did that itself.
- The `planAndSendTransactions` plugin is now deprecated, it's no longer needed as other plugins add the `plan/sendTransaction(s)` functions.

This PR is non-breaking, and everything that is deprecated works the same way it always has. Higher-level plugins like `solanaRpc` and `litesvm` have an unchanged API.

### DIFF
--- a/packages/kit-plugin-instruction-plan/README.md
+++ b/packages/kit-plugin-instruction-plan/README.md
@@ -15,9 +15,26 @@ This package provides plugins that add transaction planning and execution to you
 pnpm install @solana/kit-plugin-instruction-plan
 ```
 
+## Quick start
+
+```ts
+import { createClient } from '@solana/kit';
+import { transactionPlanner, transactionPlanSendingExecutor } from '@solana/kit-plugin-instruction-plan';
+
+const client = createClient()
+    .use(transactionPlanner(myTransactionPlanner))
+    .use(transactionPlanSendingExecutor(myTransactionPlanExecutor));
+
+// Plan without executing.
+const transactionPlan = await client.planTransactions(myInstructionPlan);
+
+// Plan and execute in one call.
+const result = await client.sendTransactions(myInstructionPlan);
+```
+
 ## `transactionPlanner` plugin
 
-The `transactionPlanner` plugin sets a custom transaction planner on the client.
+The `transactionPlanner` plugin adds `planTransaction` and `planTransactions` to the client, using the provided transaction planner.
 
 ### Installation
 
@@ -27,56 +44,6 @@ import { transactionPlanner } from '@solana/kit-plugin-instruction-plan';
 
 const myTransactionPlanner = createTransactionPlanner(/* ... */);
 const client = createClient().use(transactionPlanner(myTransactionPlanner));
-```
-
-### Features
-
-- `transactionPlanner`: A function that plans instructions into transaction messages.
-    ```ts
-    const transactionPlan = await client.transactionPlanner(myInstructionPlan);
-    ```
-
-## `transactionPlanExecutor` plugin
-
-The `transactionPlanExecutor` plugin sets a custom transaction plan executor on the client.
-
-### Installation
-
-```ts
-import { createClient, createTransactionPlanExecutor } from '@solana/kit';
-import { transactionPlanExecutor } from '@solana/kit-plugin-instruction-plan';
-
-const myTransactionPlanExecutor = createTransactionPlanExecutor(/* ... */);
-const client = createClient().use(transactionPlanExecutor(myTransactionPlanExecutor));
-```
-
-### Features
-
-- `transactionPlanExecutor`: A function that executes planned transactions.
-    ```ts
-    const transactionPlanResult = await client.transactionPlanExecutor(myTransactionPlan);
-    ```
-
-## `planAndSendTransactions` plugin
-
-The `planAndSendTransactions` plugin adds helper functions for planning and sending transactions. They accept transaction messages, instructions or instruction plans as input.
-
-### Installation
-
-This plugin requires both `transactionPlanner` and `transactionPlanExecutor` to be installed on the client.
-
-```ts
-import { createClient } from '@solana/kit';
-import {
-    transactionPlanner,
-    transactionPlanExecutor,
-    planAndSendTransactions,
-} from '@solana/kit-plugin-instruction-plan';
-
-const client = createClient()
-    .use(transactionPlanner(myTransactionPlanner))
-    .use(transactionPlanExecutor(myTransactionPlanExecutor))
-    .use(planAndSendTransactions());
 ```
 
 ### Features
@@ -93,6 +60,27 @@ const client = createClient()
     const transactionMessage = await client.planTransaction(myInstructionPlan);
     ```
 
+For backward compatibility this plugin also sets a `client.transactionPlanner` field, but that field is deprecated in favour of the two functions above.
+
+## `transactionPlanSendingExecutor` plugin
+
+The `transactionPlanSendingExecutor` plugin adds `sendTransaction` and `sendTransactions` to the client, using the client's planning functions and the provided transaction plan executor. Both functions accept transaction messages, instructions, instruction plans or transaction plans as input, planning the input first when it is not already a transaction plan.
+
+Planning goes through the client's `planTransaction` and `planTransactions` functions, so the `transactionPlanner` plugin must be installed first — otherwise the plugin throws when applied. Note that installing another `transactionPlanner` afterwards does not affect sending: `sendTransaction` and `sendTransactions` keep using the planning functions that were on the client when they were installed.
+
+### Installation
+
+```ts
+import { createClient } from '@solana/kit';
+import { transactionPlanner, transactionPlanSendingExecutor } from '@solana/kit-plugin-instruction-plan';
+
+const client = createClient()
+    .use(transactionPlanner(myTransactionPlanner))
+    .use(transactionPlanSendingExecutor(myTransactionPlanExecutor));
+```
+
+### Features
+
 - `sendTransactions`: Plans and executes transaction messages, instructions or instruction plans in one call.
 
     ```ts
@@ -105,9 +93,31 @@ const client = createClient()
     const transactionPlanResult = await client.sendTransaction(myInstructionPlan);
     ```
 
+For backward compatibility this plugin also sets a `client.transactionPlanExecutor` field, but that field is deprecated in favour of the two functions above.
+
+## Deprecated plugins
+
+The following plugins are still exported for backward compatibility but are deprecated.
+
+- **`transactionPlanExecutor(executor)`**: Only sets the deprecated `client.transactionPlanExecutor` field. Use `transactionPlanSendingExecutor` instead, which installs `sendTransaction` and `sendTransactions` alongside the executor.
+- **`planAndSendTransactions()`**: No longer needed. `transactionPlanner` now installs `planTransaction` and `planTransactions`, and `transactionPlanSendingExecutor` installs `sendTransaction` and `sendTransactions`.
+
+```ts
+// Before
+const client = createClient()
+    .use(transactionPlanner(myTransactionPlanner))
+    .use(transactionPlanExecutor(myTransactionPlanExecutor))
+    .use(planAndSendTransactions());
+
+// After
+const client = createClient()
+    .use(transactionPlanner(myTransactionPlanner))
+    .use(transactionPlanSendingExecutor(myTransactionPlanExecutor));
+```
+
 ## Default Planner and Executor Implementations
 
 For ready-to-use transaction planner and executor implementations, see:
 
-- [`@solana/kit-plugin-rpc`](https://www.npmjs.com/package/@solana/kit-plugin-rpc) — provides `rpcTransactionPlanner` and `rpcTransactionPlanExecutor` for RPC-based transaction planning and execution.
-- [`@solana/kit-plugin-litesvm`](https://www.npmjs.com/package/@solana/kit-plugin-litesvm) — provides `litesvmTransactionPlanner` and `litesvmTransactionPlanExecutor` for LiteSVM-based transaction planning and execution.
+- [`@solana/kit-plugin-rpc`](https://www.npmjs.com/package/@solana/kit-plugin-rpc) — provides the `rpcTransactionPlanner` and `rpcTransactionPlanSendingExecutor` plugins for RPC-based transaction planning and execution.
+- [`@solana/kit-plugin-litesvm`](https://www.npmjs.com/package/@solana/kit-plugin-litesvm) — provides the `litesvmTransactionPlanner` and `litesvmTransactionPlanSendingExecutor` plugins for LiteSVM-based transaction planning and execution.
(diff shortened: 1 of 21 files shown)


mcintyre94:
* **#361** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/361?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/361?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR reworks how planning/sending functions land on the client: `transactionPlanner` now installs `planTransaction(s)` directly, and a new `transactionPlanSendingExecutor` plugin installs `sendTransaction(s)` from an explicit `{ executor, planner }` pair. The rpc/litesvm packages gain `create*TransactionPlanner` / `create*TransactionPlanExecutor` factories that build a planner/executor from a client without installing anything, and the old plugins (`rpcTransactionPlanner`, `litesvmTransactionPlanExecutor`, `planAndSendTransactions`, etc.) are deprecated but delegate to the new code paths — so there's a single implementation everywhere.

This looks good to me — approving in spirit; submitting as a comment review since I don't have approve permissions on this repo.

## What I like

- **No logic duplication in deprecated paths.** `rpcTransactionPlanner` is now literally `transactionPlanner(createRpcTransactionPlanner(client, config))(client)`, and `planAndSendTransactions` composes the same `getTransactionPlanningFunctions` / `getTransactionSendingFunctions` helpers via lazy wrappers that read `client.transactionPlanner` / `client.transactionPlanExecutor` at call time — preserving the old late-binding behavior exactly.
- **Order-independence is a deliberate, documented design.** The explicit `planner` in `TransactionPlanSendingExecutorConfig` avoids install-order coupling, and both the docblocks and README spell out the consequence (installing a different `transactionPlanner` later doesn't affect `sendTransaction(s)`). The tests cover this directly (`uses its own planner rather than client.planTransactions`, `works when installed before the transactionPlanner plugin`).
- **Test coverage is strong**: the parameterized `WIRINGS` block ensures the deprecated `planAndSendTransactions` wiring keeps behaving identically to the new one, the new factories get plan-time `client.payer` read tests, and the litesvm `TContext` preservation is checked with `expectTypeOf`.
- **Changeset** (minor × 3 packages) matches the repo's semver policy for new features, and the README updates keep pace with the API changes per CONTRIBUTING.

- **Deprecated planner plugins now do more than before**: `rpcTransactionPlanner()` / `litesvmTransactionPlanner()` used to only set `client.transactionPlanner`; they now also install `planTransaction(s)`. This is additive and covered by the changeset's first sentence, but it's the one behavioral change to existing exports — worth being aware of if any consumer has its own `planTransaction` property on the client (extendClient would overwrite it).
- **`TContext` isn't surfaced on `sendTransactions` results.** `ClientWithTransactionSending` (kit-side) isn't generic, so the litesvm `transactionMetadata` context is only typed on the deprecated `transactionPlanExecutor` field, not on the result of `client.sendTransactions(...)`. Same as before this PR, so nothing to change here — but if the deprecated field is eventually removed, threading `TContext` through `ClientWithTransactionSending` in kit would be needed to keep that metadata typed.
- A couple of docstring/error-message strings still reference the deprecated plugin names from inside the new factories — see inline comments (the same applies to `createLitesvmTransactionPlanner`'s error message and the "before using this plugin" wording in both executor factories' errors).

packages/kit-plugin-rpc/src/transaction-planner.ts
```diff
@@ -44,39 +45,82 @@ import {
...
+    config: TransactionPlannerConfig = {},
+): TransactionPlanner {
+    if (config.version === 1) {
+        // The v1 transaction path is defined at the type level for forward
+        // compatibility, but `createTransactionMessage` cannot yet build v1
+        // messages, so we fail loudly rather than silently misbehave.
+        throw new Error(
+            'Version 1 transactions are not yet supported by `rpcTransactionPlanner`. ' +
+                'Use version 0 or legacy transactions for now.',
+        );
```
trevor-cortex, line 73: Nit: this error is now thrown from `createRpcTransactionPlanner` (including when called directly, not via the deprecated plugin), but the message still names `rpcTransactionPlanner`. Consider updating it to the factory name so users of the new API aren't pointed at a deprecated symbol. Same applies to `createLitesvmTransactionPlanner` in the litesvm package, and — in a similar vein — the "before using this plugin" wording in the error messages of `createRpcTransactionPlanExecutor` and `createLitesvmTransactionPlanExecutor`, which are factories rather than plugins now.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -122,68 +76,171 @@ export function rpcTransactionPlanExecutor(
...
+     *
+     * When `true` (default), any resource limit that is still provisory or
+     * unset is estimated via a simulation and replaced with the estimated
+     * value (with a compute unit buffer applied via
+     * `getComputeUnitLimitFromEstimate`). When `false`, no estimation
+     * simulation is performed and the message is sent with exactly the
+     * resource limits it already carries.
+     *
+     * This should match the `estimateResourceLimits` option passed to
+     * {@link rpcTransactionPlanner}, which controls whether provisory limits
```
trevor-cortex, line 139: Nit: now that this config type primarily backs `createRpcTransactionPlanExecutor`, this `{@link rpcTransactionPlanner}` reference points users at the deprecated plugin. Suggest referencing `createRpcTransactionPlanner` instead (the `solanaRpc` sync note still holds either way).

packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -18,26 +18,128 @@ import {
...
+ * const result = await client.sendTransactions(myInstructionPlan);
+ * ```
+ *
+ * @see {@link transactionPlanner}
+ * @see {@link TransactionPlanSendingExecutorConfig}
+ */
+export function transactionPlanSendingExecutor<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+>(config: TransactionPlanSendingExecutorConfig<TContext>) {
+    return <T extends object>(client: T) => {
```
lorisleiva: Instead of requesting the actual planner in this executor plugin, I was more thinking of using plugin dependencies to rely on the `planTransaction(s)` functions instead. Wdyt?

```suggestion
    return <T extends ClientWithTransactionPlanning>(client: T) => {
```

mcintyre94: I mentioned that as the alternative, but I think it probably is better/simpler. I didn't do it initially to make it more flexible if you have multiple planners, but having one `client.planTransaction(s)` isn't changing so I think it probably makes sense. If we need a version that customises the planner we can figure that out later.

lorisleiva: Yeah unlike the executor which can be used as a "sender" or a pure "signer", the planner doesn't really have that duality (we only have one set of `planTransaction(s)` functions). If you really wanted two separate planners you'd need custom functions anyway to distinguish them and by that point you're a very advanced user IMO so likely out-of-scope for these plugins.

mcintyre94:
Refactored so that the new `transactionPlanSendingExecutor` plugin uses the `planTransaction(s)` functions from the client. This simplifies everything else, the planner and executor no longer need to be constructed outside of the `client` pipe.

The `rpcTransactionPlanner` etc functions in rpc/litesvm are no longer replaced with `create*` functions. Instead they are refactored to construct the planner (using the input client), and then pass it to the `transactionPlanner` function. So instead of writing the planner they construct to the client themselves, they call `transactionPlanner` to do that, which also generates `planTransaction(s)` for them. But the API is unchanged.

Everything is still backward compatible. But the migration is now much simpler: `transactionPlanExecutor(...) becomes transactionPlanSendingExecutor(...)`, and `planAndSendTransactions()` can be removed.

@trevor-cortex

trevor-cortex:
Took another pass over the refactor — this is a clear improvement over the previous revision. Reading the planner off the client makes the migration story trivial (`transactionPlanExecutor(...)` → `transactionPlanSendingExecutor(...)`, drop `planAndSendTransactions()`), and keeping `rpcTransactionPlanner`/`litesvmTransactionPlanner` as plugins instead of `create*` factories means the rpc/litesvm public APIs are genuinely untouched. It also makes both of my earlier inline nits moot: the v1 error messages and the `{@link rpcTransactionPlanner}` doc references now point at symbols that are staying non-deprecated.

Things I checked and liked:

- **Apply-time planner snapshot**: `transactionPlanSendingExecutor` captures the client's planning functions when applied, so a later `transactionPlanner(...)` re-install doesn't retroactively change sending. That's a defensible choice and — importantly — it's pinned by the *"keeps using the planner installed at the time it was applied"* test, so the semantics are documented and won't drift accidentally. The deprecated `planAndSendTransactions` shim effectively has the same snapshot behavior (it closes over the client it was applied to), so there's no behavioral divergence between old and new wiring.
- **`TContext` threading**: the generic now flows through `transactionPlanSendingExecutor` and the deprecated `transactionPlanExecutor` plugin, so litesvm's `transactionMetadata` context survives on the deprecated field. That resolves the type-level gap I flagged last time.
- **Deprecated paths still share one implementation**: `rpcTransactionPlanExecutor`/`litesvmTransactionPlanExecutor` delegate to the same `createExecutor` as the new plugins, and the new tests confirm they only set the field without requiring a planner or adding `sendTransaction(s)`.
- The changeset accurately describes the new shape, including the ordering requirement on the sending executors.

Two very minor notes, neither blocking:

1. **Docblock phrasing**: *"Since sending plans through the client's planning functions, …"* (in both `rpcTransactionPlanSendingExecutor` and `litesvmTransactionPlanSendingExecutor`) reads as a garden-path sentence — "plans" lands as a noun before you realize it's the verb. Something like *"Since sending goes through the client's planning functions, …"* would read more smoothly.
2. **Planner requirement is type-only**: `transactionPlanSendingExecutor` enforces `ClientWithTransactionPlanning` at the type level, but a JS consumer (or a type-asserted client) that skips the planner won't find out until `sendTransaction` throws a `TypeError` deep inside. The svm/rpc guards set a precedent for friendly install-time errors — a tiny `if (!client.planTransactions) throw ...` in the core plugin would match that pattern. Entirely optional; the TS story covers the common case.

Overall this lands well — nice iteration.

mcintyre94:
> Docblock phrasing: "Since sending plans through the client's planning functions, …" (in both rpcTransactionPlanSendingExecutor and litesvmTransactionPlanSendingExecutor) reads as a garden-path sentence — "plans" lands as a noun before you realize it's the verb. Something like "Since sending goes through the client's planning functions, …" would read more smoothly.

Updated

> Planner requirement is type-only: transactionPlanSendingExecutor enforces ClientWithTransactionPlanning at the type level, but a JS consumer (or a type-asserted client) that skips the planner won't find out until sendTransaction throws a TypeError deep inside. The svm/rpc guards set a precedent for friendly install-time errors — a tiny if (!client.planTransactions) throw ... in the core plugin would match that pattern. Entirely optional; the TS story covers the common case.

Added this check

packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -18,26 +18,114 @@ import {
...
+export function transactionPlanSendingExecutor<
+    TContext extends TransactionPlanResultContext = TransactionPlanResultContext,
+>(transactionPlanExecutor: TransactionPlanExecutor<TContext>) {
+    return <T extends ClientWithTransactionPlanning>(client: T) => {
+        if (!client.planTransaction || !client.planTransactions) {
+            throw new Error(
+                'Transaction planning functions are required on the client to send transactions. ' +
+                    'Please add a transaction planner plugin to your client before using this plugin.',
+            );
+        }
```
lorisleiva, line 119: Random idea: could be nice to have an `isX`/`assertIsX` pair for each of these `ClientWithY` types in `@solana/plugin-interfaces` such that they use consistent `SolanaErrors` and help plugin developers guard against invalid runtime clients with just a one-liner.

packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -119,13 +255,23 @@ function getTransactionPlanningAndSendingFunctions(client: {
...
+    return { planTransaction, planTransactions };
+}
+
+function getTransactionSendingFunctions({
+    client,
+    executor,
+}: {
+    client: ClientWithTransactionPlanning;
+    executor: TransactionPlanExecutor;
+}): ClientWithTransactionSending {
```
lorisleiva: super nit: Using positional arguments here (`getTransactionSendingFunctions(client, executor)`) is slightly more consistent with the planning helper (`getTransactionPlanningFunctions(planner)`).

packages/kit-plugin-instruction-plan/README.md
```diff
@@ -15,9 +15,26 @@ This package provides plugins that add transaction planning and execution to you
 pnpm install @solana/kit-plugin-instruction-plan
 ```
 
+## Quick start
```
lorisleiva, line 18: Nice!

packages/kit-plugin-litesvm/src/litesvm.ts
```diff
@@ -40,15 +39,13 @@ export type LiteSvmConfig = {
...
-        return pipe(
+    return <T extends ClientWithPayer>(client: T) =>
+        pipe(
             client,
             litesvmConnection(),
             litesvmAirdrop(),
             litesvmGetMinimumBalance(),
             litesvmTransactionPlanner(config.transactionConfig),
-            litesvmTransactionPlanExecutor(),
-            planAndSendTransactions(),
```
lorisleiva, line 51: I love that we no longer need to set this up separately and each planner/executor is responsible for its own set of helpers.

lorisleiva (approved):
Thank you for the thorough refactoring! This is soooo much better than what we have now. 🤤

packages/kit-plugin-instruction-plan/src/core.ts
mcintyre94: Good idea, opened https://github.com/anza-xyz/kit/issues/1909 in Kit

PR #369, opened 2026-08-14
Upgrade to Kit v7.1.0 and return result contexts from executors

mcintyre94:
- Updates the `@solana/kit` dependency to v7.1.0

- Kit v7.1.0 deprecates returning a `Signature` or a `Transaction` from an executor's `executeTransactionMessage` callback in favour of returning the context that the successful result should carry. We refactor our callbacks to return the context. For rpc this is transaction + signature, litesvm also includes transactionMetadata.

- This should avoid any breaking changes when kit v8 ships.


mcintyre94:
* **#369** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
```diff
@@ -121,7 +122,11 @@ function createExecutor(client: {
             if (isFailedTransaction(result)) {
                 throw getSolanaErrorFromLiteSvmFailure(result);
             }
-            return signedTransaction;
+            return {
+                signature: getSignatureFromTransaction(signedTransaction),
+                transaction: signedTransaction,
+                transactionMetadata: result,
+            };
```
trevor-cortex: Optional nit: `transactionMetadata` is now provided twice on the success path — once via the `context.transactionMetadata = result` mutation a few lines up, and once here in the returned context. The mutation is still needed so the *failure* path carries the metadata before the throw, but you could move it inside the `isFailedTransaction` branch and let this return be the single source of truth for successes. Purely stylistic — fine as is if you prefer keeping the mutation unconditional.

trevor-cortex (commented):
This PR bumps `@solana/kit` to v7.1.0 across the workspace (root override, examples, and all five packages' peer dependencies) and migrates the two sending executors off the deprecated pattern of returning a `Signature`/`Transaction` from `executeTransactionMessage`, returning the success result context instead — `{ signature, transaction }` for the RPC executor, plus `transactionMetadata` for the LiteSVM one. Both executors gain a test asserting the success context carries a `signature` matching the returned `transaction`.

This looks clean — LGTM (would approve, but I don't have approve permissions on this repo):

- The changeset uses `minor` bumps, which is right for a raised peer-dependency floor under this repo's pre-1.0 policy (minors treated as breaking), and consumer-visible behaviour is otherwise unchanged.
- Both executors correctly keep their imperative `context.*` mutations for the failure path (LiteSVM sets `context.transaction`/`context.transactionMetadata` before the `isFailedTransaction` throw; RPC sets `context.transaction` in the pipe before `sendAndConfirmTransaction` can reject), so failed results still carry the partial context.
- In the RPC executor, deriving the signature *before* `sendAndConfirmTransaction` is a nice touch — the signature is intrinsic to the signed transaction, so there's no correctness concern with computing it early.
- No README changes needed since the documented context fields are unchanged.

For subsequent reviewers: the `pnpm-lock.yaml` churn is the usual duplication of `@solana/*` granular packages at 7.1.0 alongside remaining 7.0.0 entries (e.g. `@solana/react@7.0.0` in kit-plugin-wallet still resolves its own kit subtree) — nothing surprising, but worth a skim to confirm no unexpected major-version drift snuck in.

One optional nit inline on the LiteSVM executor.

packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
lorisleiva: There is some inconsistency between the `context` argument and the returned context:
- The returned context doesn't contain the updated message (which is different to the `TransactionPlan`'s planned message).
- The `context` argument doesn't contain the transaction signature (which we're no longer extracting from the `transaction` or it's about to change on the next major I believe).

I wonder if we're not better off returning the `context` itself to make sure they are the same even if that means we need to cast it.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -247,12 +248,13 @@ function createExecutor(
...
             );
             assertIsTransactionWithBlockhashLifetime(signedTransaction);
+            const signature = getSignatureFromTransaction(signedTransaction);
             await sendAndConfirmTransaction(signedTransaction, {
                 commitment: 'confirmed',
                 skipPreflight: skipPreflight || didSimulateToEstimate,
                 ...executorConfig,
             });
-            return signedTransaction;
+            return { signature, transaction: signedTransaction };
```
lorisleiva: Same inconsistency here:
- `context` argument is missing `signature`.
- returned context is missing `message`.

mcintyre94:
I've refactored this to better prepare for the Kit v8 migration. We now define a type `SendContext`, which precisely types what we return. Kit already allows us to parameterise the context in `createTransactionPlanExecutor` and `TransactionPlanExecutor`, so we can define and use this type now. As this is our send context it includes a `Signature`, so we comply with the current context type requirement.

Note that Kit doesn't currently parameterise the `client.sendTransaction(s)` method though, so the type doesn't get all the way through to a caller yet.

I've also refactored to use `return context as SendContext`, ie cast the mutated context instead of returning new data. Implementations that choose to return, and that specify their own context as we now do, will have type checking and it won't be possible to miss a field like `message` as my earlier version of this PR did. The cast isn't currently necessary because Kit types the input context as `SendContext`, but in v8 it'll be typed as `Partial<SendContext>` and a cast (or returning something fresh matching the context type) will be required.

@trevor-cortex

trevor-cortex:
Thanks for the ping — I've read through the refactored executors. This is a nice improvement over the previous version: the exported `SendContext` types are well documented (including the "only guaranteed on success" caveat and the LiteSVM metadata union rationale), and returning the mutated context also neatly resolves the duplication nit from my earlier review.

Two observations, neither blocking:

1. **The cast is weaker than the comment suggests.** `return context as SendContext` won't actually catch a missing field once Kit v8 types the input as `Partial<SendContext>` — TypeScript happily permits `Partial<T> as T`, so if someone later removes the `context.message = tx` assignment, this still compiles. What you *do* get from parameterising with `SendContext` is per-assignment type checking, which is real value. But if the goal is completeness checking on the return, the only pattern that provides it is building a fresh object literal (from locals, not from the `Partial` context) — e.g. `return { message, signature, transaction } satisfies SendContext`. That's a structural change you may reasonably not want given the imperative context-mutation pattern here, but worth knowing the cast alone isn't the safety net. A short code comment on the cast noting this would help future editors.

2. **Both packages export a type named `SendContext`.** Since litesvm and rpc are alternative backends, consumers importing both in one file (e.g. test suites that exercise both) will need import aliasing. Fine as-is — just flagging it as a deliberate choice worth being aware of. If it ever bothers you, `RpcSendContext` / `LitesvmSendContext` would disambiguate at the cost of symmetry.

On the "Kit doesn't parameterise `client.sendTransaction(s)` yet" point — agreed, nothing to do here until Kit threads the context type through. The `satisfies TransactionPlanExecutorConfig<SendContext>` guards are a good belt-and-braces addition in the meantime.

Overall the refactor looks sound to me — assignments all happen before the return on the success path in both executors, and the failure paths still carry whatever partial context was recorded before the throw.

mcintyre94:
Renamed the `SendContext` to `RpcSendContext` and `LiteSvmSendContext` since they're exported and were a bit ambiguous.

lorisleiva (approved):
Perfect, thanks!

PR #370, opened 2026-08-14
feat(wallet): expose reconnecting account

### LINKED ISSUE #350: Wallet plugin: Expose the wallet that is being reconnected to in state
When we initialise the wallet plugin it (optionally but by default) goes through a reconnect process to the previously used wallet, persisted to storage. This is exposed to apps through a `status: reconnecting`, but apps don't have access to the wallet that we are attempting to reconnect to until we have connected. On the first connection the app has no way to render this wallet during the reconnect, and on subsequent connections it needs to maintain the last connected wallet state itself to display it. Having this is useful to display the previous state with a disabled/loading indicator while reconnecting, for example after changing the chain. 

We could improve this DX by exposing the wallet that we are connecting to while in the `reconnecting` state. This would mean that the app doesn't need to track it itself, and would have access to it in the initial connection (if a wallet is persisted in storage).

This would only work if the wallet exposes the account in `wallet.accounts` before the reconnect, but they typically do/should. 
We'd add `reconnectingTo: UiWalletAccount | null`. We would not expose a signer.

rajanpanth:
Expose the persisted wallet account through getState().reconnectingTo while silent reconnection is in progress, without exposing a signer before the connection succeeds. The field stays null when the wallet has not registered or has not exposed the saved account, and is cleared whenever reconnecting settles.

This updates the public state documentation, adds a minor changeset, and covers browser, SSR, delayed registration, successful reconnect, and readiness transitions.

Validation:
- pnpm --dir packages/kit-plugin-wallet test:unit (187 passed)
- pnpm --dir packages/kit-plugin-wallet test:types
- pnpm --dir packages/kit-plugin-wallet build
- targeted oxlint and oxfmt --check

Closes #350

### DIFF
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -169,6 +169,8 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
 
     function updateState(updates: Partial<WalletStoreState>): void {
         const prev = state;
+        // Maintain the invariant: `reconnectingTo` is only populated when status === 'reconnecting'.
+        // If a future call site passes an explicit `reconnectingTo` with a non-reconnecting status, this clears it.
         const nextUpdates =
             updates.status !== undefined && updates.status !== 'reconnecting'
                 ? { ...updates, reconnectingTo: null }
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -57,6 +57,7 @@ export type WalletStore = WalletNamespace & {
 type WalletStoreState = {
     account: UiWalletAccount | null;
     connectedWallet: UiWallet | null;
+    reconnectingTo: UiWalletAccount | null;
     signer: WalletSigner | null;
     status: WalletStatus;
     wallets: readonly UiWallet[];
@@ -71,6 +72,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
     if (!__BROWSER__ || __REACTNATIVE__) {
         const ssrSnapshot: WalletState = Object.freeze({
             connected: null,
+            reconnectingTo: null,
             status: 'pending' as const,
             wallets: Object.freeze([]) as readonly UiWallet[],
         });
@@ -97,6 +99,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
     let state: WalletStoreState = {
         account: null,
         connectedWallet: null,
+        reconnectingTo: null,
         signer: null,
         status: 'pending',
         wallets: [],
@@ -149,6 +152,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
                           wallet: s.connectedWallet,
                       })
                     : null,
+            reconnectingTo: s.status === 'reconnecting' ? s.reconnectingTo : null,
             status: s.status,
             wallets: s.wallets,
         });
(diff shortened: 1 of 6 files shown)


packages/kit-plugin-wallet/src/store.ts
```diff
@@ -165,7 +169,11 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
 
     function updateState(updates: Partial<WalletStoreState>): void {
         const prev = state;
-        state = { ...state, ...updates };
+        const nextUpdates =
+            updates.status !== undefined && updates.status !== 'reconnecting'
+                ? { ...updates, reconnectingTo: null }
+                : updates;
+        state = { ...state, ...nextUpdates };
```
mcintyre94, line 178: I think this is a good idea, it keeps the internal state consistent. But it's a little bit strange having field-specific logic here, could you add a comment?

packages/kit-plugin-wallet/test/store.test.ts
```diff
@@ -2296,6 +2298,32 @@ describe.skipIf(!__BROWSER__)('store auto-connect (browser)', () => {
         expect(setItem).not.toHaveBeenCalled();
     });
 
+    it('exposes the saved account while a silent reconnect is in flight', async () => {
```
mcintyre94, line 2301: This test LGTM. Could you also add one for where the `mockWallet` has empty accounts (or does not include the saved account), and check that `reconnectingTo` is null in that case?

mcintyre94 (changes requested):
Nice thank you! Just a couple of small comments but this looks great :)

mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/store.ts
trevor-cortex, line 178: Nit: if a future call site ever passes an explicit `reconnectingTo` together with a non-reconnecting `status`, the auto-clear silently overrides it. That's arguably the right precedence (the invariant "no reconnectingTo outside `'reconnecting'`" wins), but a one-line comment stating that invariant here would save the next person from tracing why their explicit value got dropped. No call site hits this today, so feel free to leave as-is.

packages/kit-plugin-wallet/test/store.test.ts
```diff
@@ -2391,6 +2419,7 @@ describe.skipIf(!__BROWSER__)('store late wallet registration (browser)', () =>
 
         // Wallet not registered yet — should be reconnecting.
         expect(store.getState().status).toBe('reconnecting');
+        expect(store.getState().reconnectingTo).toBeNull();
```
trevor-cortex, line 2447: Optional coverage suggestion: this test asserts `reconnectingTo` is `null` before the wallet registers and (further down) after connect resolves, but never asserts the intermediate state where late registration triggers `attemptSilentReconnect` and `reconnectingTo` flips to the saved account while connect is still pending. Holding the connect promise open with `Promise.withResolvers` (as the in-flight test above does) would let you pin that transition down. Fine as a follow-up.

trevor-cortex (commented):
This exposes the persisted account as `getState().reconnectingTo` while a silent reconnect is in flight, without leaking a signer before the connection settles. LGTM — approving wasn't permitted from this account, so treat this as an approval.

The implementation is clean and defensive:

- The snapshot derives `reconnectingTo` guarded on `status === 'reconnecting'`, **and** `updateState` auto-clears the internal field whenever status transitions away from `'reconnecting'` — so a stale account can't leak through either path.
- The explicit `reconnectingTo: null` in the wallet-not-yet-registered path is necessary (the auto-clear only fires when a status update moves away from `'reconnecting'`), and it's correctly there.
- Change detection now includes `reconnectingTo`, so subscribers are notified when only this field changes mid-episode (e.g. late registration → account found).
- Docs (README + JSDoc), minor changeset, and the SSR snapshot are all in sync with the new field, per the repo's checklist.

Two tiny nits inline — neither blocks. For subsequent reviewers: the main thing to sanity-check is the `updateState` clearing semantics interacting with future call sites (see inline note); the current call sites are all consistent.

rajanpanth:
Thanks @mcintyre94 and @trevor-cortex for the review! I have added the invariant comment to updateState and included both suggested tests for intermediate reconnectingTo states. Pushed!

mcintyre94 (approved):
This looks great, thank you! I've also tested locally on the kit react example app and it seems to behave exactly as I'd hoped. Thanks! :)

PR #371, opened 2026-08-14
Fix dataSlice being ignored by LiteSVM getAccountInfo and getMultipleAccounts

latent-9:
The LiteSVM RPC's `getProgramAccounts` honors the `dataSlice` config option, but `getAccountInfo` and `getMultipleAccounts` silently ignored it and always returned the full account data. Their config type only listed `encoding`, and `convertMaybeEncodedAccount` never forwarded a slice to the existing `encodeAccountBase64` helper.

This threads `dataSlice` through both methods so they return the requested byte range while preserving the account's full `space`, matching `getProgramAccounts` and the Solana JSON-RPC API. Adds a test for each method covering the slice-with-full-space behavior.

### DIFF
--- a/.changeset/hip-deserts-switch.md
+++ b/.changeset/hip-deserts-switch.md
@@ -1,0 +1,5 @@
+---
+'@solana/kit-plugin-litesvm': patch
+---
+
+Fix `getAccountInfo` and `getMultipleAccounts` on the LiteSVM RPC silently ignoring the `dataSlice` config option. Both now return the requested byte range while preserving the account's full `space`, matching `getProgramAccounts` and the Solana JSON-RPC API.
--- a/packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
+++ b/packages/kit-plugin-litesvm/src/litesvm-to-rpc.ts
@@ -106,15 +106,16 @@ export function createRpcFromSvm(svm: LiteSVM): Rpc<LiteSvmRpcApi> {
     };
     const convertMaybeEncodedAccount = (
         account: MaybeEncodedAccount,
+        dataSlice?: DataSlice,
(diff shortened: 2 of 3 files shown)


lorisleiva:
@trevor-cortex

trevor-cortex (commented):
This PR fixes `getAccountInfo` and `getMultipleAccounts` on the LiteSVM RPC silently ignoring `dataSlice`, by threading the config option through `convertMaybeEncodedAccount` into the existing `encodeAccountBase64` helper (already used for slicing by `getProgramAccounts`). Both methods now return the sliced byte range while preserving the account's full `space`, and a test covers each method.

Verified:

- The slicing path reuses the exact same `sliceData`/`encodeAccountBase64` code path as `getProgramAccounts`, so behavior is consistent across all three methods.
- Test expectations are correct (`[20, 30]` → `FB4=`, `[60, 70]` → `PEY=`), and both tests assert `space: 4n` is preserved alongside the 2-byte slice, which matches the Solana JSON-RPC API semantics.
- Changeset present with a `patch` bump, appropriate for a bug fix.
- No README changes needed — the README doesn't document per-method config options for the LiteSVM RPC subset.

Notes for subsequent reviewers: the public-facing types were already correct (`GetAccountInfoApi` / `GetMultipleAccountsApi` from `@solana/kit` include `dataSlice`), so this was purely a runtime bug — the widened config types here only affect the internal handler signatures. One optional thought, not blocking: an edge-case test for an out-of-range slice (e.g. `offset` beyond the data length) would pin down the clamping behavior of `sliceData`, but since that helper predates this PR and is shared with `getProgramAccounts`, it's fine as a follow-up.

LGTM — would approve, but I don't have approval permissions on this repo.

lorisleiva (approved):
Thanks!

PR #383, opened 2026-08-21
Support version 1 transactions in the transaction planners

mcintyre94:
Kit v8 can now build version 1 transaction messages, so `litesvmTransactionPlanner` and `rpcTransactionPlanner` no longer throw when configured with `version: 1`. The planners build v1 messages, write `priorityFeeLamports` to the v1 resource header when configured, and the RPC planner reserves provisory compute unit and loaded accounts data size limits for the executor to estimate. The default version remains 0.

### DIFF
--- a/packages/kit-plugin-litesvm/README.md
+++ b/packages/kit-plugin-litesvm/README.md
@@ -157,14 +157,17 @@ const transactionPlan = await client.planTransactions(myInstructionPlan);
 
 All options are provided via a `TransactionPlannerConfig` object. Its shape is discriminated by the transaction `version`.
 
-For legacy and version 0 transactions:
+- For legacy and version 0 transactions:
 
-- `version`: The transaction message version to use. Accepts `0` or `'legacy'`. Defaults to `0`.
-- `microLamportsPerComputeUnit`: The priority fee in micro-lamports per compute unit, added as a `setComputeUnitPrice` instruction. Defaults to no priority fees.
+    - `version`: The transaction message version to use. Accepts `0` or `'legacy'`. Defaults to `0`.
+    - `microLamportsPerComputeUnit`: The priority fee in micro-lamports per compute unit, added as a `setComputeUnitPrice` instruction. Defaults to no priority fees.
+
+- For version 1 transactions:
+
+    - `version`: Set to `1` to create version 1 transaction messages.
+    - `priorityFeeLamports`: The total priority fee in lamports, written to the version 1 resource header. Defaults to no priority fees.
 
 Unlike the RPC planner, the LiteSVM planner does not estimate resource limits, since LiteSVM executes transactions locally without a simulation-based estimation step.
-
-Version 1 transactions are defined for forward compatibility but are not yet buildable by `@solana/kit`; passing `version: 1` currently throws. When available, version 1 will accept `priorityFeeLamports` (a flat total in lamports) instead of `microLamportsPerComputeUnit`.
 
 ## `litesvmTransactionPlanSendingExecutor` plugin
 
--- a/packages/kit-plugin-litesvm/src/transaction-planner.ts
+++ b/packages/kit-plugin-litesvm/src/transaction-planner.ts
@@ -7,6 +7,7 @@ import {
     pipe,
     setTransactionMessageComputeUnitPrice,
     setTransactionMessageFeePayerSigner,
+    setTransactionMessagePriorityFeeLamports,
     TransactionPlanner,
 } from '@solana/kit';
 import { transactionPlanner } from '@solana/kit-plugin-instruction-plan';
@@ -27,7 +28,6 @@ import { transactionPlanner } from '@solana/kit-plugin-instruction-plan';
  *
  * @param config - Optional configuration for the planner.
  * @returns A plugin that adds `client.planTransaction` and `client.planTransactions`.
- * @throws If `config.version` is `1`, which `@solana/kit` cannot yet build.
  *
  * @example
  * ```ts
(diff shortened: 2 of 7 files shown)


mcintyre94:
* **#383** <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anza-xyz/kit-plugins/383?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
## Summary

This PR turns the version 1 stub branch in `litesvmTransactionPlanner` and `rpcTransactionPlanner` into a working implementation now that Kit v8 can build v1 transaction messages. Instead of throwing, the planners now build v1 messages with the client's payer, optionally write `priorityFeeLamports` to the v1 resource header, and — for the RPC planner — reserve provisory compute unit and loaded accounts data size limits when `estimateResourceLimits` is enabled (the default). The obsolete `@throws` docs and `@remarks` on `TransactionPlannerConfigV1` are cleaned up, READMEs are updated to document the live options, and a minor changeset covers both packages.

The implementation looks correct to me — this would be an approval if I had the repository access to submit one:

- The v1 branches mirror the existing legacy/v0 branches faithfully, swapping `setTransactionMessageComputeUnitPrice` for `setTransactionMessagePriorityFeeLamports` and keeping the lazy `client.payer` read at plan time.
- The pipe order in the RPC planner (fill provisory limits, then set the priority fee) is the safe ordering — the explicit priority fee is applied last so it can't be clobbered by the provisory fill.
- Config defaults are unchanged (`version` defaults to 0, `estimateResourceLimits` defaults to `true`), so this is purely additive.
- Test coverage is thorough on both packages: v1 message creation, priority fee default/configured, and (RPC only) provisory limits on/off.

Keeping the v1 branch as a separate, explicit block rather than merging it with the v0 path reads well given the discriminated config union — the narrowing on `config.version === 1` is what makes `priorityFeeLamports` accessible, so I wouldn't try to consolidate.

## Things to verify

Two things worth double-checking before merge, both outside the diff itself:

1. **Kit dependency range**: the planners now import `setTransactionMessagePriorityFeeLamports` (and the tests import `getTransactionMessagePriorityFeeLamports` / `getTransactionMessageLoadedAccountsDataSizeLimit`) from `@solana/kit`. If either package's `@solana/kit` dependency/peer range still admits versions that predate these exports, consumers on the lower bound of the range would hit import failures. Worth confirming the ranges require the Kit v8 release that shipped these APIs.
2. **Executor-side support**: the RPC planner now reserves a provisory loaded accounts data size limit "for the executor to estimate", and the LiteSVM planner emits v1 messages for the LiteSVM executor to send. Neither executor is touched in this PR, so it's worth confirming that `rpcTransactionPlanSendingExecutor` already estimates the loaded accounts data size limit for v1 messages and that the LiteSVM send path handles v1 compilation — otherwise the planners are writing headers nothing downstream fulfils yet.

Neither is blocking assuming they were validated as part of the Kit v8 upgrade, but a quick confirmation in the PR thread would be useful for the record.

mcintyre94:
> Kit dependency range

Is now >8, so we're good

> Executor-side support

RPC executor uses Kit's `estimateAndSetResourceLimits`, which includes data size limit

packages/kit-plugin-litesvm/src/transaction-planner.ts
```diff
@@ -56,13 +56,19 @@ export function litesvmTransactionPlanner(config: TransactionPlannerConfig = {})
...
-        throw new Error(
-            'Version 1 transactions are not yet supported by `litesvmTransactionPlanner`. ' +
-                'Use version 0 or legacy transactions for now.',
-        );
+        const priorityFeeLamports = config.priorityFeeLamports;
+        return createTransactionPlanner({
+            createTransactionMessage: () => {
+                return pipe(
+                    createTransactionMessage({ version: 1 }),
+                    tx => setTransactionMessageFeePayerSigner(client.payer, tx),
```
lorisleiva: super nit: all this is the same as the v0/legacy path. I know it might be tricky because of the config union but it might be slightly cleaner to branch a bit later? Something like

```ts
return createTransactionPlanner({
    createTransactionMessage: () => {
        return pipe(
            createTransactionMessage({ version }),
            tx => setTransactionMessageFeePayerSigner(client.payer, tx),
            tx => setVersionSpecificDetails(tx),
        );
    },
});
```

Not a big deal if TypeScript makes this awkward and it's cleaner to branch our the whole return statement just thought it's probably worth a shot. ☺️

packages/kit-plugin-litesvm/README.md
```diff
@@ -164,7 +164,10 @@ For legacy and version 0 transactions:
 
 Unlike the RPC planner, the LiteSVM planner does not estimate resource limits, since LiteSVM executes transactions locally without a simulation-based estimation step.
```
lorisleiva: nit: I'd probably push that at the end of the section so the v0 and v1 lists and next to each other. I'd also make them sublists so they look more like config references. Something like:

- For legacy and version 0 transactions:
  - `versions`: [...]
  - `microLamportsPerComputeUnit`: [...]
- For version 1 transactions:
  - `versions`: [...]
  - `priorityFeeLamports`: [...]

packages/kit-plugin-rpc/src/transaction-planner.ts
```diff
@@ -64,13 +64,21 @@ export function rpcTransactionPlanner(config: TransactionPlannerConfig = {}) {
...
-            'Version 1 transactions are not yet supported by `rpcTransactionPlanner`. ' +
-                'Use version 0 or legacy transactions for now.',
-        );
+        const estimateResourceLimits = config.estimateResourceLimits ?? true;
+        const priorityFeeLamports = config.priorityFeeLamports;
+        return createTransactionPlanner({
+            createTransactionMessage: () => {
+                return pipe(
+                    createTransactionMessage({ version: 1 }),
+                    tx => setTransactionMessageFeePayerSigner(client.payer, tx),
```
lorisleiva: Same nit here

packages/kit-plugin-rpc/README.md
```diff
@@ -275,7 +275,11 @@ For legacy and version 0 transactions:
 - `microLamportsPerComputeUnit`: The priority fee in micro-lamports per compute unit, added as a `setComputeUnitPrice` instruction. Defaults to no priority fees.
 - `estimateResourceLimits`: Whether to estimate and set resource limits by simulating before sending. Set to `false` to skip estimation and reserve no provisory limits, which is useful for transactions close to the message size limit. Defaults to `true`.
 
-Version 1 transactions are defined for forward compatibility but are not yet buildable by `@solana/kit`; passing `version: 1` currently throws. When available, version 1 will accept `priorityFeeLamports` (a flat total in lamports) instead of `microLamportsPerComputeUnit`, alongside the shared `estimateResourceLimits` option.
+For version 1 transactions:
```
lorisleiva: Same nit here on the nested list.

lorisleiva (approved):
Nice, thank you! Just a few optional nits.

packages/kit-plugin-litesvm/src/transaction-planner.ts
mcintyre94: I tried a few things and I think branching at this point ends up fighting the type system too much because it can't reconcile that tx version and config version match in a clean way.

I ended up splitting it into two functions that take a type narrowed config, eg `createV1Planner(client: ClientWithPayer, config: TransactionPlannerConfigV1)`

And then `createPlanner` just branches on `config` version: `return config.version === 1 ? createV1Planner(client, config) : createLegacyPlanner(client, config);`

I think that's probably the cleanest option, and also means it'll be easy to support more versions in future.

PR #390, opened 2026-08-24
Add transaction plan signing executor

mcintyre94:
Add a transaction plan signing plugin that installs Kit's `ClientWithTransactionSigning` methods to `kit-plugin-instruction-plan`.

This is the same idea as the sending one, it takes an executor and installs its functionality as the `signTransaction(s)` functions. This executor will be provided by the rpc and LiteSVM plugins calling it.

The common transaction execution flow is shared with the sending plugin. This refactor isn't really necessary for only two implementations, but I think it makes sense because it demonstrates how we split the logic between the input executor and these client functions, and I think any other addition should use the same pattern.

Note: Also added AGENTS.md as a symlink from CLAUDE.md, same as we do in Kit

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/packages/kit-plugin-instruction-plan/src/core.ts
+++ b/packages/kit-plugin-instruction-plan/src/core.ts
@@ -5,17 +5,24 @@ import {
     CanceledSingleTransactionPlanResult,
     ClientWithTransactionPlanning,
     ClientWithTransactionSending,
+    ClientWithTransactionSigning,
     createFailedToSendTransactionError,
     createFailedToSendTransactionsError,
+    createFailedToSignTransactionError,
+    createFailedToSignTransactionsError,
     extendClient,
     FailedSingleTransactionPlanResult,
+    InstructionPlanInput,
     isSolanaError,
     isTransactionPlan,
     parseInstructionOrTransactionPlanInput,
     parseInstructionPlanInput,
     singleTransactionPlan,
     SOLANA_ERROR__INSTRUCTION_PLANS__FAILED_TO_EXECUTE_TRANSACTION_PLAN,
+    SuccessfulSingleTransactionPlanResult,
+    SingleTransactionPlan,
     TransactionPlanExecutor,
+    TransactionPlanInput,
     TransactionPlanner,
     TransactionPlanResult,
     TransactionPlanResultContext,
@@ -107,6 +114,7 @@ export function transactionPlanner(transactionPlanner: TransactionPlanner) {
  * ```
  *
  * @see {@link transactionPlanner}
+ * @see {@link transactionPlanSigningExecutor}
  */
 export function transactionPlanSendingExecutor<
     TContext extends TransactionPlanResultContext = TransactionPlanResultContextWithSignature,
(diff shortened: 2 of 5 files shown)


mcintyre94:
@trevor-cortex

packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -263,7 +324,42 @@ function getTransactionSendingFunctions<TContext extends TransactionPlanResultCo
...
+        createTransactionError: (
+            result: CanceledSingleTransactionPlanResult<TContext> | FailedSingleTransactionPlanResult<TContext>,
+            abortReason?: unknown,
+        ) => Error;
+        createTransactionsError: (result: TransactionPlanResult<TContext>, abortReason?: unknown) => Error;
+    },
+): {
+    executeTransaction: ClientWithTransactionSigning<TContext>['signTransaction'];
+    executeTransactions: ClientWithTransactionSigning<TContext>['signTransactions'];
+} {
```
trevor-cortex: Nit: the shared helper's return type (and the local `const` annotations below) borrow their shape from `ClientWithTransactionSigning`, which reads a bit oddly when the helper is also the implementation behind `getTransactionSendingFunctions`. It compiles because the sending and signing function types are structurally identical today, but that's an implicit coupling — if the two client interfaces ever diverge, this helper would silently keep the signing shape.

Consider declaring the shared function types once, locally, and referencing those from both call sites, e.g.:

```ts
type ExecuteTransactionsFunction<TContext extends TransactionPlanResultContext> =
    ClientWithTransactionSigning<TContext>['signTransactions'] &
    ClientWithTransactionSending<TContext>['sendTransactions'];
```

or simply a standalone type alias not derived from either client interface. Entirely non-blocking — the current code is correct.

AGENTS.md
```diff
@@ -0,0 +1 @@
+CLAUDE.md
```
trevor-cortex, line 1: Assuming this is committed as a symlink to `CLAUDE.md` (mode `120000`) — the patch rendering is consistent with that, but I can't see the file mode from here. Worth double-checking with `git ls-files -s AGENTS.md`: if it landed as a regular file whose content is the literal text `CLAUDE.md`, tools that look for `AGENTS.md` would just see a filename string instead of the instructions.

trevor-cortex (commented):
## Summary

This PR adds a `transactionPlanSigningExecutor` plugin that installs Kit's `ClientWithTransactionSigning` methods (`signTransaction` / `signTransactions`) on clients, mirroring the existing `transactionPlanSendingExecutor`. The common plan-then-execute flow is extracted into a shared `getTransactionExecutionFunctions` helper parameterized by error factories, so both plugins share identical planning, abort-signal, and error re-wrapping behavior. Also includes an `AGENTS.md` pointing at `CLAUDE.md`, README docs, a minor changeset, and a solid test suite.

The refactor is a nice touch — beyond deduplication, it actually fixes a small typing gap in the sending path: the error re-wrapping previously cast to untyped `TransactionPlanResult` and now preserves `TContext` through `TransactionPlanResult<TContext>` and the single-result variants.

This looks good to me — I'd approve, but I don't have repository access to do so formally.

## Notes / questions (all non-blocking)

- **`TContext` default parity**: `transactionPlanSendingExecutor` defaults `TContext` to `TransactionPlanResultContextWithSignature`, while the signing plugin has no default. That seems deliberate since signing contexts are implementation-specific (signed transactions vs. signatures), but if Kit exposes an equivalent `...WithSignedTransaction` context type, it might be worth defaulting to it for symmetry. If the asymmetry is intentional, feel free to ignore.
- **Optional test**: the `signTransaction` success/single-result assertion path (`assertIsSuccessfulSingleTransactionPlanResult`) is only exercised through the sending WIRINGS suite. Since the code is now shared that's arguably covered, but one test asserting `signTransaction` rejects on a multi-transaction result would pin the behavior the README documents ("Should the provided input result in multiple transactions, an error will be thrown") against future divergence of the two paths.
- **Docblock symmetry**: the signing docblock has `@see {@link transactionPlanSendingExecutor}`; a reciprocal `@see` on the sending plugin's docblock would help discoverability.

## For subsequent reviewers

- The `expectTypeOf` assertion in the "preserving the executor context type" test is doing real work here — it verifies `TContext` inference flows from the executor through to the result context, which is the main point of the plugin per the PR description.
- Changeset is a `minor` bump for a new feature, generated via the CLI per repo convention. README and docblocks follow the existing structure of the sending plugin sections.

mcintyre94:
> Optional test: the signTransaction success/single-result assertion path

Good idea, added

> Docblock symmetry

Added, both point to each other now

> Nit: the shared helper's return type

Moved these to their own types `ExecuteTransaction(s)Function` defined here. It's a bit of a complex type and added some more imports, but I think it makes more sense than re-using either type for both sign and send.

lorisleiva (approved):
Nice, thank you!

PR #395, opened 2026-08-25
Add the `rpcTransactionPlanSigningExecutor` plugin

mcintyre94:
This PR adds `client.signTransaction(s)` for the `solanaRpc` client, using a new `rpcTransactionPlanSigningExecutor` plugin. A follow-up similar PR will add the same for LiteSVM.

This plugin performs most of the same functionality as the `rpcTransactionPlanSendingExecutor`, it attaches a newly fetched blockhash lifetime (memoized, the same for each transaction), performs resource estimation using simulation, and partially signs the transaction(s). It does not attempt to send the transactions, so that the app is able to do that separately.

It does not use `createTransactionPlanExecutor`, and instead ignores sequential dependencies in the input transaction plans: since we are not executing the transactions these dependencies are irrelevant. All leaf single transaction plans are handled in parallel, but with a maximum concurrency (the same approach used for send). This means that signing doesn't fan out infinitely by default, since some signers may require user input or calling services with rate limiting.

The returned `TransactionPlanResult` has the same shape as the input, so sequential dependencies/divisibility/etc are maintained. 

Noted in the readme, but because we are not executing transactions, if there are genuine sequential dependencies then simulation of later transactions will fail, and `estimateResourceLimits: false` will need to be used. This is a disadvantage compared to send, which is able to ensure that sequential dependencies are sent + confirmed first.

Unlike the send path, signing uses `partiallySignTransactionMessageWithSigners` and does not require all signers to be present. The returned transaction plans have a context containing a transaction message, a transaction, and optionally a signature. If the fee payer has signed then `signature` is populated, else it is omitted.

Note that like the send path, the lifetime is replaced with the fresh blockhash, replacing for example a durable nonce lifetime. We could skip our lifetime replacement for sign, and allow an existing durable nonce lifetime to remain. But I think given that we know there are plans to replace and deprecate durable nonces, it's ok not to do that here.

Also note that the executor concurrency is separate from send, meaning that an app might be concurrently handling sign and send operations. I think this is generally the right approach, it doesn't make sense to queue signing before a send + confirm that might take much longer. But it does mean that if an RPC has a limited requests per second, this is not necessarily an effective way to manage that. This isn't new, you could always have RPC requests outside of the `client.sendTransaction(s)` flow causing it to hit a limit. It may be useful to add a `requestsPerSecond` to `solanaRpc` which rate-limits the `client.rpc` to make this easier and robust, I'll open an issue.

### DIFF
--- a/packages/kit-plugin-rpc/src/transaction-plan-executor.ts
+++ b/packages/kit-plugin-rpc/src/transaction-plan-executor.ts
@@ -4,6 +4,7 @@ import {
     ClientWithRpcSubscriptions,
     ClientWithTransactionPlanning,
     createTransactionPlanExecutor,
+    createTransactionPlanExecutorWithConcurrentLeaves,
     estimateAndSetResourceLimitsFactory,
     estimateResourceLimitsFactory,
     GetEpochInfoApi,
@@ -11,6 +12,7 @@ import {
     GetSignatureStatusesApi,
     getSignatureFromTransaction,
     isSolanaError,
+    partiallySignTransactionMessageWithSigners,
     pipe,
     ResourceLimitsEstimate,
     SendableTransaction,
@@ -29,9 +31,16 @@ import {
     TransactionMessageWithFeePayer,
     TransactionPlanExecutor,
     TransactionPlanExecutorConfig,
+    TransactionWithinSizeLimit,
     TransactionWithLifetime,
+    Base64EncodedWireTransaction,
+    getBase64EncodedWireTransaction,
 } from '@solana/kit';
-import { transactionPlanExecutor, transactionPlanSendingExecutor } from '@solana/kit-plugin-instruction-plan';
+import {
+    transactionPlanExecutor,
+    transactionPlanSendingExecutor,
+    transactionPlanSigningExecutor,
+} from '@solana/kit-plugin-instruction-plan';
 
 /**
  * A plugin that adds `sendTransaction` and `sendTransactions` to the client
@@ -79,7 +88,48 @@ export function rpcTransactionPlanSendingExecutor(config: RpcTransactionPlanExec
             ClientWithTransactionPlanning,
     >(
         client: T,
-    ) => transactionPlanSendingExecutor(createExecutor(client, config))(client);
+    ) => transactionPlanSendingExecutor(createSendingExecutor(client, config))(client);
+}
+
+/**
+ * A plugin that adds `signTransaction` and `signTransactions` to the client
+ * using a transaction plan executor backed by RPC.
+ *
+ * The executor fetches a blockhash, estimates resource limits when needed, and
+ * partially signs every transaction in the plan. Transaction plan execution
+ * constraints are ignored during signing, while transaction preparation and
+ * signing respect `maxConcurrency`.
+ *
+ * Since signing goes through the client's planning functions,
+ * {@link rpcTransactionPlanner} or another `transactionPlanner` plugin must be
+ * installed first.
+ *
+ * @param config - Optional configuration for the signing executor.
+ * @returns A plugin that adds `client.signTransaction` and `client.signTransactions`.
+ * @throws If the client has no RPC or no transaction planning functions set.
+ *
+ * @example
+ * ```ts
+ * import { createClient } from '@solana/kit';
+ * import { rpcTransactionPlanSigningExecutor, rpcTransactionPlanner, solanaRpcConnection } from '@solana/kit-plugin-rpc';
+ * import { generatedPayer } from '@solana/kit-plugin-signer';
+ *
+ * const client = await createClient()
+ *     .use(solanaRpcConnection({ rpcUrl: 'https://api.mainnet-beta.solana.com' }))
+ *     .use(generatedPayer())
+ *     .use(rpcTransactionPlanner())
+ *     .use(rpcTransactionPlanSigningExecutor());
+ *
+ * const result = await client.signTransactions(myInstructionPlan);
+ * ```
+ *
+ * @see {@link rpcTransactionPlanner}
+ * @see {@link rpcTransactionPlanSendingExecutor}
+ */
+export function rpcTransactionPlanSigningExecutor(config: RpcTransactionPlanSigningExecutorConfig = {}) {
+    return <T extends ClientWithRpc<GetLatestBlockhashApi & SimulateTransactionApi> & ClientWithTransactionPlanning>(
+        client: T,
+    ) => transactionPlanSigningExecutor(createSigningExecutor(client, config))(client);
 }
 
 /**
@@ -122,7 +172,7 @@ export function rpcTransactionPlanExecutor(config: RpcTransactionPlanExecutorCon
             ClientWithRpcSubscriptions<SignatureNotificationsApi & SlotNotificationsApi>,
     >(
         client: T,
-    ) => transactionPlanExecutor(createExecutor(client, config))(client);
+    ) => transactionPlanExecutor(createSendingExecutor(client, config))(client);
 }
 
 /**
@@ -194,6 +244,33 @@ export type RpcTransactionPlanExecutorConfig = {
 };
 
 /**
+ * Configuration for {@link rpcTransactionPlanSigningExecutor}.
+ *
+ * Signing uses the same resource limit estimation options as
+ * {@link rpcTransactionPlanSendingExecutor}, but it has no preflight behavior
+ * because transactions are not sent.
+ *
+ * @see {@link rpcTransactionPlanSigningExecutor}
+ */
+export type RpcTransactionPlanSigningExecutorConfig = {
+    /**
+     * Whether to estimate and set resource limits by simulating each transaction
+     * before signing. Defaults to `true`.
+     */
+    estimateResourceLimits?: boolean;
+    /**
+     * Maps estimated compute unit consumption to the limit set on the
+     * transaction. The result is capped at 1,400,000 compute units.
+     */
+    getComputeUnitLimitFromEstimate?: (estimatedComputeUnits: number) => number;
+    /**
+     * The maximum number of transactions prepared and signed concurrently
+     * across all calls to this signing executor. Defaults to 10.
+     */
+    maxConcurrency?: number;
+};
+
+/**
  * The context carried by transaction plan results from
  * {@link rpcTransactionPlanSendingExecutor}.
  *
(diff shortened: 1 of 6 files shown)


mcintyre94:
@trevor-cortex

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -299,6 +421,125 @@ function createExecutor(
...
+                                context.message = message;
+                            }
+
+                            const transaction = await partiallySignTransactionMessageWithSigners(
+                                message,
+                                executorConfig,
+                            );
+                            context.transaction = transaction;
+
+                            // May be partially signed, only include signature if available
```
trevor-cortex: Nit: this relies on the first entry of `transaction.signatures` being the fee payer's slot (which is what makes it a safe guard for the `getSignatureFromTransaction` call below, since that function reads the same first entry). That assumption holds for Kit-compiled transactions, but it's implicit here — and the docs (`RpcSignContext`, README) specifically promise "signature present when the *fee payer* signed". Consider making the comment explicit, e.g. `// The first signature slot belongs to the fee payer; only expose a signature once it has signed.` so the coupling to the fee-payer-first invariant is visible.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -299,6 +421,125 @@ function createExecutor(
...
+    }
+
+    const shouldEstimateResourceLimits = config.estimateResourceLimits ?? true;
+    const getComputeUnitLimitFromEstimate =
+        config.getComputeUnitLimitFromEstimate ?? getDefaultComputeUnitLimitFromEstimate;
+    const estimateResourceLimits = estimateResourceLimitsFactory({ rpc: client.rpc });
+    const executeSigningTransaction = limitFunction(
+        (execute: () => Promise<SingleTransactionPlanResult<RpcSignContext>>) => execute(),
+        config.maxConcurrency ?? 10,
+    );
```
trevor-cortex: Nit: the bare `false` here is `skipPreflight`, a concept that doesn't exist on the signing path. A short inline hint would help readers, e.g. `bufferAndRecoverResourceLimits(estimateResourceLimits, getComputeUnitLimitFromEstimate, /* skipPreflight */ false)` — making it clear that failed estimation simulations always throw (and surface as failed leaf results) when signing, since there's no "send anyway for debugging" escape hatch.

trevor-cortex (commented):
This PR adds `rpcTransactionPlanSigningExecutor`, wiring `client.signTransaction(s)` into the `solanaRpc` client. The executor memoizes a single fresh blockhash per call, optionally estimates resource limits via simulation (reusing `bufferAndRecoverResourceLimits`), and partially signs each leaf with `partiallySignTransactionMessageWithSigners`, bounded by a shared `maxConcurrency` limiter. It deliberately bypasses `createTransactionPlanExecutor` and walks the plan itself, running all leaves concurrently while preserving the input plan's shape (parallel/sequential/divisibility) in the result.

The implementation is clean and the design decisions are well-reasoned and well-documented — this looks good to merge. (Posting as COMMENT since I don't have approval permissions on this repo.)

- Ignoring sequential execution constraints is correct for signing, and the README honestly documents the consequence (simulation of dependent transactions can fail → use `estimateResourceLimits: false`).
- Attempting every leaf even after a sibling fails (rather than canceling, as the sending path's non-divisible sequential semantics would) is a sensible divergence for signing and is covered by a test.
- The lazy, memoized blockhash fetch avoids an orphaned RPC request for empty plans and guarantees all leaves share one lifetime — both tested.
- Skipping `assertIsTransactionWithBlockhashLifetime` on the signed transaction is intentional, letting a modifying signer swap in e.g. a durable nonce lifetime; there's a test pinning that behavior.
- Error flow correctly funnels through `createFailedToExecuteTransactionPlanError` so the shared plugin layer rewraps it as `SOLANA_ERROR__FAILED_TO_SIGN_TRANSACTIONS`, with the full plan result attached.

Test coverage is thorough (empty plans, malformed plan kinds, unsigned fee payer, modifying signers, cross-call concurrency sharing, partial failure reporting, missing RPC), the changeset is a `minor` bump per the repo's policy, and the README/docblocks follow the existing structure.

Notes for subsequent reviewers / follow-ups:

- `executeSigningTransactionPlan` is fully generic — nothing RPC-specific beyond the context type. Since a LiteSVM signing executor is planned as a follow-up, consider hoisting this leaf-walking helper (and possibly the limiter wiring) into `@solana/kit-plugin-instruction-plan` at that point to avoid duplicating it.
- The fee-payer signature detection relies on the first entry of `transaction.signatures` being the fee payer (see inline comment) — worth a second pair of eyes to confirm that assumption holds for all signer types, including transactions returned by modifying signers.
- Minor race, consistent with existing behavior elsewhere: if a genuine signing failure lands after the abort signal fires, the leaf is classified as canceled and the underlying error is dropped. Not worth changing here, just noting it's pre-existing semantics.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
mcintyre94: This holds for all Solana transactions, it's a protocol concern not Kit

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
mcintyre94: Added the `* skipPreflight */ false` comment

For completeness, the reason we don't have the `skipPreflight` option is because it'd only be used on this call to simulate, and you can opt out of that with `estimateResourceLimits: false`.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -299,6 +421,129 @@ function createExecutor(
...
```
lorisleiva: nit: I wonder if it would be clearer if we dissolved this helper. It made it hard for me to see where concurrency was because it was hidden inside a helper that doesn't really mention that: `executeSigningTransaction`.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
```diff
@@ -299,6 +421,129 @@ function createExecutor(
...
+                            context.transaction = transaction;
+
+                            // May be partially signed, only include signature if available
+                            if (Object.values(transaction.signatures)[0]) {
+                                context.signature = getSignatureFromTransaction(transaction);
+                            }
+
+                            return context.signature
+                                ? { message, signature: context.signature, transaction }
+                                : { message, transaction };
```
lorisleiva: How do you feel about adding a base64 encoding of the transaction in the context so it's that much easier to send a signed transaction to a multisig?

lorisleiva: nit: In order to avoid branching the returned context (and adding new attribute twice), we could do something like this:

```ts
return {
  ...(context.signature ? { signature: context.signature } : {}),
  message,
  transaction,
};
```

Although here we could also just `return context as RpcSignContext`.

lorisleiva (approved):
Thanks! Just a few nits but looks good to me.

packages/kit-plugin-rpc/src/transaction-plan-executor.ts
mcintyre94: Good idea, added `transactionBase64: Base64EncodedWireTransaction`

mcintyre94:
Moving to draft for now - we're going to move the concurrent execution helper to Kit.

mcintyre94:
This PR has been refactored to upgrade to Kit 8.1.0 and use the new `createTransactionPlanExecutorWithConcurrentLeaves` function. This makes the implementation much more similar to send, as the helpers have the same structure.

mcintyre94:
@trevor-cortex

trevor-cortex (commented):
Re-review after the latest push. The delta since my last pass looks good — this remains ready to merge from my side (posting as COMMENT since I don't have approval permissions on this repo; note Loris has already approved).

**What changed since the last review:**

- The PR now includes the `@solana/kit@^8.1.0` upgrade across all packages (own changeset, `loud-masks-shake`), which unlocks the new `createTransactionPlanExecutorWithConcurrentLeaves` primitive from Kit. The hand-rolled `executeSigningTransactionPlan` leaf-walker from the previous revision is gone, replaced by that shared helper — this fully resolves my earlier follow-up note about hoisting the generic leaf-walking logic before the LiteSVM signing executor lands. The malformed-plan-kind invariant test confirms the shared helper's plan validation is wired through.
- `bufferAndRecoverResourceLimits` now takes an optional `onSimulate` instead of the signing path passing a no-op, and the signing call site carries the `/* skipPreflight */ false` hint — both prior nits addressed, and the docblock was updated accordingly.
- The blockhash memoization is now per-executor-call (`latestBlockhashPromise` scoped inside the returned executor function) while the `limitFunction` concurrency budget remains shared across calls at plugin-install time — the right split, and both behaviors are pinned by tests (`getLatestBlockhash` called once per `signTransactions` call; concurrency shared across simultaneous calls).
- README and `solanaRpc` docs updated to describe the independent sign/send concurrency budgets, matching the discussion in the PR description.

**Remaining notes (non-blocking):**

- The `loud-masks-shake` changeset ships the Kit 8.1 requirement as `minor` with a **BREAKING CHANGES** callout in the body. That matches the pre-1.0 convention these packages appear to follow (breaking changes surface as minor bumps), but worth a quick sanity check against the repo's release precedent before merge.
- The fee-payer signature detection still relies on the fee-payer-first ordering of `transaction.signatures` (raised in my first review — not re-raising, just flagging for the merger that it was a deliberate call if it was discussed in the resolved threads).

Nothing new to flag inline — the implementation and test coverage remain thorough.

mcintyre94:
This PR has been updated again to use Kit 8.2.0 and its updated version of `createTransactionPlanExecutorWithConcurrentLeaves`. This API better matches `createTransactionPlanExecutor`. As the original version of this PR (without the Kit helper) did, it now maintains partial context if the sign fails/is cancelled, better matching `client.sendTransaction(s)`.

@trevor-cortex

trevor-cortex:
Thanks for the ping @mcintyre94 — took a look at the Kit 8.2.0 update and the delta looks good.

A few things I verified:

- **Partial context on failure/cancel**: the new `executeTransactionMessage` signature (receiving `context: Partial<RpcSignContext>` and mutating it progressively) correctly restores the original PR's behavior and matches how the send path records context. Nice touch assigning `context.message` both before and after estimation, so a failed simulation still leaves the blockhash-attached message on the failed result. The `@remarks` on `RpcSignContext` documenting that failed/canceled results carry pre-failure values is exactly the right place for that caveat.
- **Abort handling**: the `executorConfig.abortSignal?.throwIfAborted()` guard before the lazy blockhash fetch prevents orphaned RPC requests for leaves that were queued behind the concurrency limit when the signal fired — and the new abort test pins that (releasing the slot and asserting no further signing/simulation calls).
- **`transactionBase64` addition** to the context is a sensible convenience for the "app sends separately" use case, and it's asserted against `getBase64EncodedWireTransaction(transaction)` in the tests.
- The changeset correctly reflects the `^8.0.0` → `^8.2.0` peer dep bump across all packages.

Nothing new to flag. The fee-payer-first signature ordering note from my earlier reviews still stands as-is — no change needed, just leaving the pointer for whoever merges.

.changeset/loud-masks-shake.md
```diff
@@ -0,0 +1,13 @@
...
+'@solana/kit-plugin-rpc': minor
+'@solana/kit-plugin-signer': minor
+'@solana/kit-plugin-wallet': minor
+---
+
+Upgrade all packages to `@solana/kit@^8.2.0`.
+
+**BREAKING CHANGES**
+
+**Kit 8.2 is now required.** Upgrade `@solana/kit` from `^8.0.0` to `^8.2.0`.
```
lorisleiva, line 13: nit: could we make this a separate prepended PR?

lorisleiva (approved):
Re-approving since changes. thank you!

PR #396, opened 2026-08-26
Add the `litesvmTransactionPlanSigningExecutor` plugin

mcintyre94:
This PR adds `client.signTransaction(s)` for the `solanaLitesvm` client, using a new `litesvmTransactionPlanSigningExecutor` plugin. This follows from the equivalent plugin for RPC: #395 

This plugin performs most of the same functionality as the `litesvmTransactionPlanSendingExecutor`, it attaches a blockhash lifetime using `client.svm`, and partially signs the transaction(s). It does not attempt to send the transactions, so that the app is able to do that separately.

As with the RPC version, it uses `createTransactionPlanExecutorWithConcurrentLeaves` which ignores sequential dependencies in the input transaction plan.

Unlike the RPC version, no concurrency limit is used, matching the behaviour of liteSVM send. 

The returned context is the same as the litesvm send one, except `signature` is optional and `transactionBase64` is added. This doesn't have the same multisig use case as RPC but can be useful for debugging, so I think it fits the litesvm use case well.

### DIFF
--- a/packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
+++ b/packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
@@ -1,7 +1,12 @@
 import {
+    Base64EncodedWireTransaction,
     ClientWithTransactionPlanning,
     createTransactionPlanExecutor,
+    createTransactionPlanExecutorWithConcurrentLeaves,
+    getBase64EncodedWireTransaction,
     getSignatureFromTransaction,
+    isFullySignedTransaction,
+    partiallySignTransactionMessageWithSigners,
     pipe,
     SendableTransaction,
     Signature,
@@ -12,9 +17,14 @@ import {
     TransactionMessageWithFeePayer,
     TransactionPlanExecutor,
     TransactionPlanExecutorConfig,
+    TransactionWithinSizeLimit,
     TransactionWithLifetime,
 } from '@solana/kit';
-import { transactionPlanExecutor, transactionPlanSendingExecutor } from '@solana/kit-plugin-instruction-plan';
+import {
+    transactionPlanExecutor,
+    transactionPlanSendingExecutor,
+    transactionPlanSigningExecutor,
+} from '@solana/kit-plugin-instruction-plan';
 import type { FailedTransactionMetadata, LiteSVM, TransactionMetadata } from 'litesvm';
 
 import { getSolanaErrorFromLiteSvmFailure, isFailedTransaction } from './transaction-error';
@@ -58,7 +68,48 @@ import { getSolanaErrorFromLiteSvmFailure, isFailedTransaction } from './transac
  */
 export function litesvmTransactionPlanSendingExecutor() {
     return <T extends ClientWithTransactionPlanning & { svm: LiteSVM }>(client: T) =>
-        transactionPlanSendingExecutor(createExecutor(client))(client);
+        transactionPlanSendingExecutor(createSendingExecutor(client))(client);
+}
+
+/**
+ * A plugin that adds `signTransaction` and `signTransactions` to the client
+ * using a transaction plan executor backed by LiteSVM.
+ *
+ * The executor sets each message's lifetime from LiteSVM, partially signs every
+ * transaction in the plan, and simulates fully signed transactions to collect
+ * transaction metadata without committing state changes. Transaction plan
+ * execution constraints are ignored during signing, so every transaction is
+ * prepared and signed concurrently. A failed simulation is recorded as
+ * {@link FailedTransactionMetadata} but does not make signing fail.
+ *
+ * Since signing goes through the client's planning functions,
+ * {@link litesvmTransactionPlanner} or another `transactionPlanner` plugin must
+ * be installed first.
+ *
+ * @returns A plugin that adds `client.signTransaction` and `client.signTransactions`.
+ * @throws If the client has no LiteSVM instance or no transaction planning functions set.
+ *
+ * @example
+ * ```ts
+ * import { createClient } from '@solana/kit';
+ * import { litesvmConnection, litesvmTransactionPlanner, litesvmTransactionPlanSigningExecutor } from '@solana/kit-plugin-litesvm';
+ * import { generatedPayer } from '@solana/kit-plugin-signer';
+ *
+ * const client = await createClient()
+ *     .use(litesvmConnection())
+ *     .use(generatedPayer())
+ *     .use(litesvmTransactionPlanner())
+ *     .use(litesvmTransactionPlanSigningExecutor());
+ *
+ * const result = await client.signTransactions(myInstructionPlan);
+ * ```
+ *
+ * @see {@link litesvmTransactionPlanner}
+ * @see {@link litesvmTransactionPlanSendingExecutor}
+ */
+export function litesvmTransactionPlanSigningExecutor() {
+    return <T extends ClientWithTransactionPlanning & { svm: LiteSVM }>(client: T) =>
+        transactionPlanSigningExecutor(createSigningExecutor(client))(client);
 }
 
 /**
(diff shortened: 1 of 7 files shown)


mcintyre94:
@trevor-cortex

trevor-cortex (commented):
This PR adds a `litesvmTransactionPlanSigningExecutor` plugin providing `client.signTransaction(s)` for LiteSVM clients, mirroring the RPC version from #395. It sets a fresh LiteSVM blockhash lifetime, partially signs each transaction, and — unlike the RPC version — simulates fully signed transactions (without committing state) to attach `TransactionMetadata` to the result context. The shared tree-traversal helper is promoted to `@solana/kit-plugin-instruction-plan` as `executeTransactionPlanLeavesConcurrently`, and the RPC executor now consumes it.

Overall this is well executed: the structure closely follows the RPC signing executor, the docblocks and READMEs are thorough, the changeset is appropriate (`minor` on both packages), and test coverage is strong — including the modifying-signer lifetime case, failed-sibling reporting, and a real LiteSVM simulation test gated behind `__NODEJS__`.

**Main design question** (inline): a failed simulation fails the leaf, which means `signTransactions` on a sequential plan with state-dependent transactions will *always* throw, since later transactions are simulated against un-mutated state. The README documents this, but it makes the executor throw for exactly the sequential-dependency case the concurrent traversal was designed to support. Worth deciding whether simulation failure should fail the signing result at all (or be opt-out).

Smaller notes:

- The lack of a concurrency limit is called out as deliberate (matching LiteSVM send), and that's reasonable for a local SVM — but note the signers themselves can still be remote (hardware/wallet signers), and the new instruction-plan README explicitly advises wrapping the callback in a limit when signers can't handle unbounded requests. Not blocking, just flagging the asymmetry with the RPC version's `maxConcurrency`.
- Now that `executeTransactionPlanLeavesConcurrently` is a public API, its docblock could clarify the contract around callbacks that throw (inline).

For subsequent reviewers: the `kit-plugin-rpc` change is a pure extraction of `executeSigningTransactionPlan` into the shared helper — I diffed it against the original and it's behavior-identical. The area needing the most scrutiny is the simulation-failure semantics discussed above.

packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
```diff
@@ -174,3 +271,82 @@ function createExecutor(client: { svm: LiteSVM }): TransactionPlanExecutor<LiteS
...
+
+                            if (isFullySignedTransaction(transaction)) {
+                                const simulationResult = client.svm.simulateTransaction(transaction);
+                                context.transactionMetadata = isFailedTransaction(simulationResult)
+                                    ? simulationResult
+                                    : simulationResult.meta();
+                                if (isFailedTransaction(simulationResult)) {
+                                    throw getSolanaErrorFromLiteSvmFailure(simulationResult);
+                                }
+                            }
```
trevor-cortex: Design question: a failed simulation throws here, turning the leaf into a failed result and making the whole `signTransactions` call throw `SOLANA_ERROR__FAILED_TO_SIGN_TRANSACTIONS` — even though signing itself succeeded.

Combined with the concurrent, dependency-ignoring traversal, this means any sequential plan whose later transactions depend on earlier state will *always* fail: each fully signed leaf is simulated against the current (un-mutated) LiteSVM state, as the README itself notes. So the executor throws for precisely the sequential-dependency case this signing flow is meant to support.

A few options worth considering:

1. Treat simulation as purely informational: record the `FailedTransactionMetadata` in the context but return a *successful* result, since the signing job succeeded. Consumers can inspect `isFailedTransaction(context.transactionMetadata)` themselves.
2. Keep the current behavior but add a config flag (e.g. `simulateTransactions?: boolean` or `throwOnSimulationFailure?: boolean`) so dependent sequential plans have an escape hatch — analogous to how the RPC version can disable `estimateResourceLimits`.

Option 1 feels most aligned with "sign, don't execute", but either would unblock the sequential case. If the current behavior is intentional, the README's "Sequential dependencies" section should say explicitly that the call will throw, not just that simulation "can fail".

packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
```diff
@@ -174,3 +271,82 @@ function createExecutor(client: { svm: LiteSVM }): TransactionPlanExecutor<LiteS
...
+
+                            return {
+                                ...(context.signature ? { signature: context.signature } : {}),
+                                message: context.message,
+                                transaction: context.transaction,
+                                transactionBase64: context.transactionBase64,
+                                transactionMetadata: context.transactionMetadata,
+                            };
```
trevor-cortex: Minor consistency nit: `signature` is conditionally spread so it's omitted when absent, but `transactionMetadata` is always assigned — so a partially signed result carries an explicit `transactionMetadata: undefined` property. `'transactionMetadata' in context` then behaves differently from `'signature' in context`, and this would also trip `exactOptionalPropertyTypes` if it's ever enabled. Consider mirroring the spread pattern:

```ts
...(context.transactionMetadata ? { transactionMetadata: context.transactionMetadata } : {}),
```

packages/kit-plugin-instruction-plan/src/core.ts
```diff
@@ -195,6 +202,62 @@ export function transactionPlanSigningExecutor<TContext extends TransactionPlanR
...
+    switch (kind) {
+        case 'single':
+            return await executeSingleTransactionPlan(transactionPlan);
+        case 'parallel':
+        case 'sequential': {
+            const plans = await Promise.all(
+                transactionPlan.plans.map(plan =>
+                    executeTransactionPlanLeavesConcurrently(plan, executeSingleTransactionPlan),
+                ),
+            );
```
trevor-cortex: Now that this is a public API, it may be worth documenting the callback error contract: if `executeSingleTransactionPlan` *throws* (rather than returning a failed `SingleTransactionPlanResult`), this `Promise.all` rejects, the sibling results are discarded, and the shape-preservation guarantee no longer holds — only the first rejection surfaces. Both in-repo callers catch inside the callback and return failed results, which is the pattern external callers should follow. A one-liner in the docblock (e.g. "the callback should return a failed result instead of throwing to preserve the result shape") would make that explicit.

packages/kit-plugin-litesvm/src/transaction-plan-executor.ts
mcintyre94: This has been updated, if simulation fails we now return a successful result with `transactionMetadata: FailedTransactionMetadata`.

packages/kit-plugin-instruction-plan/src/core.ts
mcintyre94: This is outdated, the helper is now `createTransactionPlanExecutorWithConcurrentLeaves` in Kit

mcintyre94:
As with #395 this PR has been rewritten to use `createTransactionPlanExecutorWithConcurrentLeaves`

@trevor-cortex

trevor-cortex:
Thanks for the ping @mcintyre94 — I took a pass over the rewritten version and it addresses everything from my earlier review:

- **Simulation is now informational**: failed simulations land as `FailedTransactionMetadata` on a *successful* signing result instead of failing the whole call, and simulation is gated on `isFullySignedTransaction` so partial transactions no longer hit sigverify failures. This resolves my main concern about sequential plans with state-dependent transactions always throwing. The new tests (`retains failed simulation metadata in a successful result context`, `does not fail sequential plans when their simulations fail`) cover it well, and the README/docblocks document the semantics clearly.
- **Context consistency**: `transactionMetadata` now uses conditional spread like `signature`, so the explicit-`undefined` inconsistency (and the `exactOptionalPropertyTypes` hazard) is gone — and the test even asserts `not.toHaveProperty('transactionMetadata')`. 👍
- **Concurrent-leaves helper**: with `createTransactionPlanExecutorWithConcurrentLeaves` now living in Kit rather than this repo, my note about its docblock (callbacks must return failed results rather than throw, or the shape-preservation guarantee breaks) applies upstream — might be worth carrying over there if it isn't already documented.

One small nit while re-reading: the "attempts every sequential leaf" test confirms sibling leaves still succeed when one signer rejects, which is a nice guarantee to have pinned down.

Nothing blocking from my side — this looks good to go.

lorisleiva (approved):
Thanks!

PR #402, opened 2026-08-28
docs: add kit-plugin-resilient-subscriptions to community plugins

MayurK-cmd:
Adds `kit-plugin-resilient-subscriptions` to the community plugins table.

This plugin adds resilient WebSocket subscription utilities to any Kit client, including:
- Auto-reconnect with exponential backoff and jitter
- Optional gap recovery to fetch missed state after reconnection
- Abort signal support for graceful shutdown
- Parallel recovery coordination across multiple subscriptions

Repo: https://github.com/MayurK-cmd/kit-plugin-resilient-subscriptions


mcintyre94:
This plugin references Kit APIs that don't exist so we can't add it currently

PR #403, opened 2026-08-29
Stop executing queued transactions after abort in the sending executor

mcintyre94:
When a transaction plan has more leaves than `maxConcurrency`, leaves waiting for a concurrency slot in `rpcTransactionPlanSendingExecutor` would still fetch a blockhash, estimate resource limits, sign and send once a slot freed up, even after the plan execution had been aborted. The executor has already reported those leaves at that point, so this work was orphaned: its RPC requests and signer calls served no result. The executor now checks the abort signal as soon as a queued execution acquires its slot and bails out before doing any work. 

This was identified as part of writing the RPC signing executor, pre-existing on the send executor.

### DIFF
--- a/packages/kit-plugin-rpc/test/transaction-plan-executor.test.ts
+++ b/packages/kit-plugin-rpc/test/transaction-plan-executor.test.ts
@@ -24,6 +24,7 @@ import {
     singleInstructionPlan,
     singleTransactionPlan,
     SingleTransactionPlanResult,
+    SOLANA_ERROR__FAILED_TO_SEND_TRANSACTIONS,
     SOLANA_ERROR__FAILED_TO_SIGN_TRANSACTIONS,
     SOLANA_ERROR__INVARIANT_VIOLATION__INVALID_TRANSACTION_PLAN_KIND,
     SolanaRpcApi,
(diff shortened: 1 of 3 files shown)


mcintyre94:
@trevor-cortex

packages/kit-plugin-rpc/test/transaction-plan-executor.test.ts
```diff
@@ -680,6 +682,49 @@ describe('rpcTransactionPlanSendingExecutor', () => {
...
+        abortController.abort();
+        // Fail the in-flight transaction so the overall call settles.
+        sendReleasers[0]();
+        const error = await pendingError;
+        assert(isSolanaError(error, SOLANA_ERROR__FAILED_TO_SEND_TRANSACTIONS));
+
+        // The concurrency slot held by the first transaction is now free, so the
+        // queued second one gets its turn; verify it bails out without doing any work.
+        await new Promise(resolve => setTimeout(resolve, 0));
+        await new Promise(resolve => setTimeout(resolve, 0));
```
trevor-cortex, line 746: Minor, non-blocking: these two macrotask flushes are timing-based — if the queued execution's turn ever takes more ticks (e.g. an extra microtask hop inside the executor), the negative assertions below would pass vacuously rather than actually exercising the bail-out path. Since a true "never called" can't be `waitFor`-ed, one way to make a regression fail loudly instead is to have the `getLatestBlockhash` mock throw (or record) on a second invocation, so any post-abort work trips an explicit failure regardless of tick count. Fine as-is given the comment explains the intent.

trevor-cortex (commented):
This PR fixes a real orphaned-work bug in `rpcTransactionPlanSendingExecutor`: when a plan has more leaves than `maxConcurrency`, executions queued in `limitFunction` would still fetch a blockhash, simulate, sign, and send after the plan had already been aborted — work the executor had already written off. The fix is a single well-placed `executorConfig?.abortSignal?.throwIfAborted()` at the top of the limited function, so a queued execution bails the moment it acquires its slot.

The fix looks correct to me (would approve, but I don't have approval rights on this repo):

- The check sits inside the `limitFunction` wrapper, which is exactly where the gap was — the kit-level executor passes the abort signal down via `executorConfig` to the RPC calls and signers, but nothing guarded the window between abort and a queued call acquiring its concurrency slot.
- Throwing `signal.reason` here is safe: the kit executor attached its handlers to the returned promise when it initiated the call, and (per the code comment) has already reported the leaf, so the rejection is absorbed rather than surfacing as unhandled.
- Since `rpcTransactionPlanExecutor` shares `createExecutor`, the deprecated plugin gets the fix for free.

The test is well-constructed — `maxConcurrency: 1` with two parallel leaves, holding the first send open until after abort, then verifying the queued leaf does zero work once the slot frees. The early `.catch` on `pendingError` to avoid an unhandled rejection while the abort settles is a nice touch.

Changeset is a `patch` on `@solana/kit-plugin-rpc`, which matches the repo's policy for fixes, and the random filename indicates it was CLI-generated.

Notes for subsequent reviewers:

- `AbortSignal.throwIfAborted` needs Node 17.3+/modern browsers; the React Native test environment in CI is the one place this could theoretically bite. If the three-environment test suite passes, this is a non-issue.
- The final negative assertions in the new test rely on two macrotask flushes to give the queued execution its turn (see inline note) — timing-based, but bounded and commented, so I'm fine with it.

lorisleiva (approved):

PR #414, opened 2026-09-01
Add `signOffchainMessage` wallet action

mcintyre94:
Draft for now, will hold off on merging until we have wallet support for the feature. But reviewable 

This PR adds the wallet-standard `solana:signOffchainMessage` feature as `client.wallet.signOffchainMessage`, and a `useSignOffchainMessage` react hook for it. This allows apps to ask the connected wallet to sign a given message using the offchain message format. 

The function takes as input an object `{version: 1, message: string, requiredSigners?: Address[]}`, ie the fields for a version 1 offchain message. See https://github.com/solana-foundation/SRFCs/discussions/3 

`requiredSigners` is optional, if omitted it defaults to the connected wallet address. If provided then it must include the connected wallet address. It returns a Kit `OffchainMessageEnvelope`, which will be partially signed if there are more than one `requiredSigners`.

The returned message is decoded and Kit's `assertOffchainMessageV1Equal` is called to ensure it is equivalent to the requested message. The wallet-standard feature requires the wallet to construct the message envelope. The returned signature is verified against the returned message.

These checks are different to other wallet plugin functions. For transactions we use the Kit signers API, which can return a modified transaction. We don't verify signatures for the app for transactions, but apps typically don't need to trust them and can rely on sending the transaction. The `signMessage` API has always been more limited and we don't verify its returned signature either. We designed the `signOffchainMessage` feature to be easy to verify, and apps rely on its signatures offchain by design, so I think it makes sense to have the stricter checks here.

### DIFF
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -1,9 +1,20 @@
 import {
+    type Address,
+    assertOffchainMessageV1Equal,
+    getAddressEncoder,
+    getOffchainMessageDecoder,
+    getPublicKeyFromAddress,
+    type OffchainMessageBytes,
+    type OffchainMessageEnvelope,
     type ReadonlyUint8Array,
     type SignatureBytes,
+    SOLANA_ERROR__OFFCHAIN_MESSAGE__ADDRESSES_CANNOT_SIGN_OFFCHAIN_MESSAGE,
+    SOLANA_ERROR__OFFCHAIN_MESSAGE__SIGNATURE_VERIFICATION_FAILURE,
+    SOLANA_ERROR__OFFCHAIN_MESSAGE__UNEXPECTED_VERSION,
     SOLANA_ERROR__WALLET__ACCOUNT_NOT_AVAILABLE,
     SOLANA_ERROR__WALLET__NOT_CONNECTED,
     SolanaError,
+    verifySignature,
 } from '@solana/kit';
 import { createSignerFromWalletAccount } from '@solana/wallet-account-signer';
 import {
@@ -13,6 +24,8 @@ import {
     type SolanaSignInOutput,
     SolanaSignMessage,
     type SolanaSignMessageFeature,
+    SolanaSignOffchainMessage,
+    type SolanaSignOffchainMessageFeature,
 } from '@solana/wallet-standard-features';
 import { getWallets } from '@wallet-standard/app';
 import {
@@ -37,6 +50,7 @@ import type {
     WalletNamespace,
     WalletPluginConfig,
     WalletSigner,
+    WalletSignOffchainMessageInput,
     WalletState,
     WalletStatus,
     WalletStorage,
@@ -91,6 +105,10 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
             signIn: () => Promise.reject(new SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED, { operation: 'signIn' })),
             signMessage: () =>
                 Promise.reject(new SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED, { operation: 'signMessage' })),
+            signOffchainMessage: () =>
+                Promise.reject(
+                    new SolanaError(SOLANA_ERROR__WALLET__NOT_CONNECTED, { operation: 'signOffchainMessage' }),
+                ),
             subscribe: () => () => {},
             subscribeSigner: () => () => {},
             whenReady: () => Promise.resolve(),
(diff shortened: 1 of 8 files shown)


mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -713,6 +731,77 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
...
+            });
+        }
+        assertOffchainMessageV1Equal(receivedMessage, {
+            content: message,
+            requiredSignatories: requiredSigners.map(address => ({ address })),
+            version,
+        });
+
+        // Verify the signature returned by the wallet
+        const content = output.signedOffchainMessage as OffchainMessageBytes;
```
trevor-cortex, line 804: The envelope returned to the app aliases the `Uint8Array` instances the wallet handed back. A misbehaving wallet (or anything else retaining a reference to `output`) can mutate `signedOffchainMessage` or `signature` *after* this verification runs, leaving the app holding an envelope whose `content`/`signatures` no longer match what was verified — a classic TOCTOU gap, and this function's whole premise is not trusting the wallet.

Consider copying both buffers up front and doing all decode/verify work (and envelope construction) against the copies:

```ts
const content = Uint8Array.from(output.signedOffchainMessage) as unknown as OffchainMessageBytes;
const signature = Uint8Array.from(output.signature) as SignatureBytes;
```

(`Object.freeze` on the envelope doesn't help here — it freezes the object's properties, not the typed arrays' contents.)

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -713,6 +731,77 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
...
+                unexpectedAddresses: [signerAddress],
+            });
+        }
+
+        const addressEncoder = getAddressEncoder();
+        const [output] = await signOffchainMessageFeature.signOffchainMessage({
+            account,
+            message,
+            messageVersion: version,
+            requiredSigners: requiredSigners.map(signer => addressEncoder.encode(signer)),
```
trevor-cortex, line 784: Optional: the `solana:signOffchainMessage` feature exposes `supportedMessageVersions` (the test mock declares it), but the store never consults it. If a future wallet doesn't support v1 (or, later, an app requests a version the wallet lacks), this call fails with an arbitrary wallet-side error instead of something actionable. Checking `signOffchainMessageFeature.supportedMessageVersions.includes(version)` up front and throwing a Kit `SolanaError` would fail faster and more clearly — and slots in naturally when a v2 union member is added to `WalletSignOffchainMessageInput`.

trevor-cortex (commented):
This PR adds the wallet-standard `solana:signOffchainMessage` feature as `client.wallet.signOffchainMessage` plus a `useSignOffchainMessage` React hook. Unlike `signMessage`, the action treats the wallet's response as untrusted: it decodes the returned bytes, asserts they encode exactly the requested v1 message and signer set via `assertOffchainMessageV1Equal`, verifies the signature against the connected account's key, and resolves with a Kit `OffchainMessageEnvelope` (partially signed when `requiredSigners` has more than one entry).

The implementation is solid and the stricter verification posture is well justified given apps rely on these signatures offchain — I'd approve if I had the repo access to do so. Things I checked:

- **Verification ordering** is right: decode → explicit version check (`UNEXPECTED_VERSION` before the v1-only assert can run) → content/signatory equality → signature verification. Each failure mode maps to a distinct, documented Kit error, and every one has test coverage.
- **Pre-flight `requiredSigners` check** correctly rejects an explicit list omitting the connected account before the wallet is ever called, and the default of `[connected account]` matches the docs.
- **Signature verification against the connected account's *address*** (via `getPublicKeyFromAddress`) rather than the account object's `publicKey` field is the right call — the address is what the store's state vouches for.
- **Tests** are genuinely strong: real Ed25519 key pairs so the verify path runs against genuine signatures, round-trip compatibility with `verifyOffchainMessageEnvelope`, and a case per error code including a v0 envelope for the version mismatch.
- Docblocks, README section, and changeset (minor, feature) all follow repo conventions.

Two suggestions inline, both in the spirit of the PR's own "don't trust the wallet" design:

1. The returned envelope aliases the `Uint8Array` instances the wallet handed back, so they can be mutated after verification (TOCTOU). Copying before verifying would close this.
2. Consider consulting the feature's `supportedMessageVersions` before dispatching, to fail fast with a Kit error rather than an arbitrary wallet-side one.

For subsequent reviewers: the main thing to scrutinize is the verification block in `store.ts` (~L769–801) — specifically that the signature is verified over the *full* returned envelope bytes (preamble + body, per the SRFC) and that the equality assert covers everything an app would rely on. The rest is mechanical plumbing (hook, types, docs). Note the PR is a draft pending wallet-side support for the feature, per the description.

packages/kit-plugin-wallet/src/store.ts
mcintyre94: I think this is fine for now, we should add a check when we have multiple versions. Currently the feature only uses v1 messages.

mcintyre94:
This PR is rebased/stacked on #415 and has the same fix to get the wallet's account to pass to the feature

PR #431, opened 2026-09-10
Add `supportedTransactionVersions` to the wallet plugin's connected state

mcintyre94:
This PR adds `connected.supportedTransactionVersions`, so that apps can easily read the supported transaction versions for the connected wallet account. This mirrors wallet-adapter which includes this on the adapter object. As wallets have differing levels of support for v1 this is more useful so I think worth adding to our plugin too.

This also raises `@solana/wallet-standard-features` to `^1.5.0`, adding `1` to the union type.

The value is derived from wallet features in a similar way to what wallet-adapter does.  See https://github.com/anza-xyz/wallet-standard/blob/d7c8537a889b67c66c1df2c438ff194defcd4790/packages/wallet-adapter/base/src/adapter.ts#L202-L205

Updated: We use the intersection of `supportedTransactionVersions` of whichever sign features are available, `signTransactions` or `signAndSendTransactions`. This is conservative, in practice wallets should support the same versions for these features. But since our signer exposes both, this is the safest option. Unlike wallet-adapter we do not null the field in support is only `['legacy']`. If wallets report weird combinations then we already have `getWalletAccountFeature` as an escape hatch to get the value from each feature. 

The derivation is scoped to the connected account rather than the wallet, which diverges from wallet-adapter. This means that it correctly updates if a wallet switches for example from a signing account to a read-only one.

### DIFF
--- a/packages/kit-plugin-wallet/src/store.ts
+++ b/packages/kit-plugin-wallet/src/store.ts
@@ -7,12 +7,17 @@ import {
 } from '@solana/kit';
 import { createSignerFromWalletAccount } from '@solana/wallet-account-signer';
 import {
+    SolanaSignAndSendTransaction,
+    type SolanaSignAndSendTransactionFeature,
     SolanaSignIn,
     type SolanaSignInFeature,
     type SolanaSignInInput,
     type SolanaSignInOutput,
     SolanaSignMessage,
     type SolanaSignMessageFeature,
+    SolanaSignTransaction,
+    type SolanaSignTransactionFeature,
+    type SolanaTransactionVersion,
 } from '@solana/wallet-standard-features';
 import { getWallets } from '@wallet-standard/app';
 import {
@@ -64,8 +69,22 @@ type WalletStoreState = {
     reconnectingTo: UiWalletAccount | null;
     signer: WalletSigner | null;
     status: WalletStatus;
+    supportedTransactionVersions: ReadonlySet<SolanaTransactionVersion>;
     wallets: readonly UiWallet[];
 };
+
+// The subset of the state derived from a single wallet account's chains and
+// features. Derived together so the two can never disagree about the account
+// they describe.
+type AccountSignerState = Pick<WalletStoreState, 'signer' | 'supportedTransactionVersions'>;
+
+// Shared empty set for accounts that support no transaction versions, so the
+// value stays referentially stable while no account (or a read-only one) is
+// active.
+const NO_TRANSACTION_VERSIONS: ReadonlySet<SolanaTransactionVersion> = new Set();
+
+// The features that carry `supportedTransactionVersions`.
+const TRANSACTION_SIGNING_FEATURES = [SolanaSignAndSendTransaction, SolanaSignTransaction] as const;
 
 // -- Store ------------------------------------------------------------------
 
@@ -106,6 +125,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
         reconnectingTo: null,
         signer: null,
         status: 'pending',
+        supportedTransactionVersions: NO_TRANSACTION_VERSIONS,
         wallets: [],
     };
 
@@ -153,6 +173,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
                     ? Object.freeze({
                           account: s.account,
                           signer: s.signer,
+                          supportedTransactionVersions: s.supportedTransactionVersions,
                           wallet: s.connectedWallet,
                       })
                     : null,
@@ -204,6 +225,7 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
             state.reconnectingTo !== prev.reconnectingTo ||
             state.status !== prev.status ||
             state.signer !== prev.signer ||
+            state.supportedTransactionVersions !== prev.supportedTransactionVersions ||
             state.wallets !== prev.wallets
         ) {
             snapshot = deriveSnapshot(state);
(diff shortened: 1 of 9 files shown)


mcintyre94:
@trevor-cortex

packages/kit-plugin-wallet/src/store.ts
```diff
@@ -243,6 +265,38 @@ export function createWalletStore(config: WalletPluginConfig): WalletStore {
...
+    //
+    // Read account-scoped rather than wallet-scoped for the same reason the
+    // signer is: an account may expose a narrower feature set than its wallet,
+    // and the registry regenerates the account handle precisely when that set
+    // changes. Feature resolution can still throw for a malformed or stale
+    // wallet, which degrades to the empty set rather than failing the
+    // connection over a metadata read.
+    function getSupportedTransactionVersions(account: UiWalletAccount): ReadonlySet<SolanaTransactionVersion> {
+        const featureName = TRANSACTION_SIGNING_FEATURES.find(name => account.features.includes(name));
+        if (!featureName) return NO_TRANSACTION_VERSIONS;
```
trevor-cortex: Non-blocking design question. `createSignerFromWalletAccount` returns a signer with *both* `modifyAndSignTransactions` (backed by `solana:signTransaction`) and `signAndSendTransactions` (backed by `solana:signAndSendTransaction`) when the account has both features, and Kit's helpers pick one depending on the call site (`signTransactionMessageWithSigners` vs `signAndSendTransactionMessageWithSigners`). So if a wallet advertises, say, `['legacy', 0]` on sign-and-send but `['legacy']` on sign-only, this reports v0 as supported even though the `modifyAndSignTransactions` path the app might actually take doesn't support it.

Wallet-adapter gets away with the same precedence because its adapter routes everything through one path. Two alternatives if you want the value to be safe for whichever path the signer takes:

- Intersect the two sets when both features are present (conservative — "versions every signing path supports").
- Keep the current rule but call out in the docblock/README that the set reflects the sign-and-send feature when present.

In practice wallets almost always advertise the same versions on both, so this is an edge case. Happy with the current behaviour if it's a deliberate choice — mainly flagging so it's decided rather than inherited.

trevor-cortex (commented):
## Summary

Adds `connected.supportedTransactionVersions: ReadonlySet<SolanaTransactionVersion>` to the wallet plugin state. The value is derived per-account alongside the signer (new `deriveAccountSignerState`), preferring `solana:signAndSendTransaction` and falling back to `solana:signTransaction`, with a shared frozen empty set for accounts that have neither or whose feature lookup throws. Bumps `@solana/wallet-standard-features` to `^1.5.0` so the version union includes `1`. Includes changeset (minor — correct for a new public field), README section, docblocks on the type and the React hook, and browser tests for precedence, legacy-only, empty, referential stability, regeneration, and the malformed-wallet degradation path.

**Verdict: LGTM.** (Submitting as a comment since I can't formally approve on this repo.) One non-blocking design question inline.

## What I checked

- Every path that nulls `connectedWallet`/`account` also resets `supportedTransactionVersions` to `NO_TRANSACTION_VERSIONS` (`disconnectLocally`, the unregister handler). `revertToPreviousConnectionOrDisconnect` and the no-target `disconnect` branch are covered transitively since they only touch `status` when a connection is intact. The SSR stub needs no change since `connected` is `null` there.
- `updateState`'s snapshot-invalidation condition includes the new field, and `reconcileActiveAccount` still returns `{}` on same-reference, so the referential-stability test is meaningful.
- Redundant-looking pre-check `account.features.includes(name)` before `getWalletAccountFeature` is fine — it keeps the common "no signing feature" case off the exception path.
- `new Set(undefined)` doesn't throw, so a wallet that omits `supportedTransactionVersions` entirely degrades to an empty set without hitting the `catch`. Good.

## One design point (non-blocking, inline)

The precedence rule mirrors wallet-adapter, but Kit's `createSignerFromWalletAccount` exposes *both* `modifyAndSignTransactions` (via `solana:signTransaction`) and `signAndSendTransactions` (via `solana:signAndSendTransaction`) when both features are present, and which one Kit's signer-aware helpers invoke depends on the call site. So when the two features advertise different version sets, the reported value describes only one of the two paths the exposed `signer` can take. See inline comment for options; I'm fine with the current choice given the wallet-adapter precedent as long as it's a deliberate decision.

- The lockfile shows `@solana/wallet-account-signer@8.2.0` already resolving `@solana/wallet-standard-features` to 1.5.0, so the dep bump doesn't introduce a duplicate copy.
- The PR description highlights `selectAccount` switching from a signing account to a read-only one as a motivating case. That path goes through the shared `deriveAccountSignerState`, so it's covered by construction, but there's no test that exercises `selectAccount` directly for this field. Optional to add one.
- The `_setup.ts` mock for `solana:signTransaction`/`solana:signAndSendTransaction` uses module-level mutable state (`supportedTransactionVersions`, `transactionFeatureFailure`) reset in `beforeEach`, consistent with how the other mocks in that file already work.

packages/kit-plugin-wallet/src/store.ts
mcintyre94: Changed this to use an intersection. I think in practice wallets will be reporting the same values so this is an edge case as mentioned.

Using `signAndSend` can already under-report if `sign` has a wider value, so intersection seems reasonable to me.

Apps have `getWalletAccountFeature` as an escape hatch if a wallet is reporting weird combinations here, and if that happens then we can re-assess or look at surfacing values from multiple features here.

lorisleiva (approved):
Nice addition, thanks!

